Issue #6, opened 2015-06-05
Published version of electron-rebuild isn't compatible with the latest babel-core


hongrich:
The current published version of electron-rebuild (0.2.1) isn't compatible with the latest releases of babel-core. This seems to be related to changes in regenerator https://github.com/babel/babel/issues/1685

Recompiling the source files and publishing a new version should fix the problem. You can reproduce this issue by downloading the released tarball and running the tests in this repo (but not regenerating the source files) against regenerator@0.8.28 (a dependency of babel-core)

Here's the output from running the test (I removed the part where it regenerates the lib files, so it can just used the published version):
```
% npm test

> electron-rebuild@0.2.1 test /Users/rich/src/electron-rebuild/dist/package
> npm run compile && mocha test-dist/*


> electron-rebuild@0.2.1 compile /Users/rich/src/electron-rebuild/dist/package
> babel --stage 0 -d test-dist/ test/

test/main.js -> test-dist/main.js
test/promisify.js -> test-dist/promisify.js
test/spawn.js -> test-dist/spawn.js
test/support.js -> test-dist/support.js


  installNodeHeaders
    1) installs node headers for 0.25.2

  rebuildNativeModules
    2) Rebuilds native modules against 0.22.0

  shouldRebuildNativeModules
    ✓ should always return true most of the time maybe

  promisify
    ✓ should handle fs.stat
    ✓ should map all of fs

  spawn
    ✓ should work with ls
    ✓ should fail when the path is completely bogus


  5 passing (62ms)
  2 failing

  1) installNodeHeaders installs node headers for 0.25.2:
     ENOENT, stat '/Users/rich/src/electron-rebuild/dist/package/test-dist/testheaders/.node-gyp/0.25.2/common.gypi'
  Error: ENOENT, stat 'test-dist/testheaders/.node-gyp/0.25.2/common.gypi'

  2) rebuildNativeModules Rebuilds native modules against 0.22.0:
     ENOENT, stat '/Users/rich/src/electron-rebuild/dist/package/test-dist/testheaders/.node-gyp/0.22.0/common.gypi'
  Error: ENOENT, stat 'test-dist/testheaders/.node-gyp/0.22.0/common.gypi'
```

paulcbetts:
:-/  Not super excited that this is Possible™

paulcbetts:
Also, thanks a ton for reporting this, I was busy tearing my hair out trying to figure out why this doesn't work anymore!

[...]

PR #7, opened 2015-06-05
Fix checking for installed headers

hongrich:
Fix checking for installed headers to actually pass in nodeVersion as the first param. This also includes a test.


hongrich:
I discovered this while looking into #6. I couldn't get `to.throw()` to work in the test, so I just followed the same pattern you used in the promisify test

paulcbetts:
Thanks!

PR #11, opened 2015-07-19
Add option for specifying electron-prebuilt path

benogle:
This will allow an electron-cli module to use the electron that is specified by the app.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -25,7 +25,7 @@ const argv = require('yargs')
   .epilog('Copyright 2015')
   .argv;
 
-if (!argv.e){
+if (!argv.e) {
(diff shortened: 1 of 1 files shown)


src/cli.js
```diff
@@ -20,20 +20,23 @@ const argv = require('yargs')
   .alias('a', 'arch')
   .describe('m', 'The path to the node_modules directory to rebuild')
   .alias('m', 'module-dir')
+  .describe('e', 'The path to electron-prebuilt')
+  .alias('e', 'electron-prebuilt-dir')
```
benogle: Not sure about the name here, or if there is something better we can specify.

paulcbetts:
Into it, thanks @benogle!

PR #28, opened 2015-10-13
feature/quick-rebuilds

cdaringe:
# problem statement
- rebuilds are slow for large projects
- rebuilds re-build content that doesn't _need to be rebuilt_

# solution
- maintain a cache of modules that have been built
- rebuild only those modules that have not been built since last `electron-rebuild`
    - because we cannot control users from doing something like `rm -rf node_modules/some_module`, our cache is at risk of being invalided.  Therefore, cache `node_modules/some_folder`'s  `stat.ctime`, and assert that the cached `ctime` === the folders current `ctime`.  If they're not ===, fallback safely by just rebuilding `some_module`

### DIFF
--- a/test/main.js
+++ b/test/main.js
@@ -1,45 +1,34 @@
 import _ from './support';
 import path from 'path';
+import pkgMock from 'mock-npm-install';
 import promisify from '../lib/promisify';
+import { mkdir, rmdir } from './utils/fileio.js';
 const fs = promisify(require('fs'));
-const rimraf = promisify(require('rimraf'));
 const cp = promisify(require('ncp').ncp);
 
 import {installNodeHeaders, rebuildNativeModules, shouldRebuildNativeModules} from '../lib/main.js';
 
 describe('installNodeHeaders', function() {
   this.timeout(30*1000);
-  
+
   it('installs node headers for 0.25.2', async () => {
     let targetHeaderDir = path.join(__dirname, 'testheaders');
+    await rmdir(targetHeaderDir);
+    await mkdir(targetHeaderDir);
 
-    try {
-      if (await fs.stat(targetHeaderDir)) {
-        await rimraf(targetHeaderDir);
-      }  
-    } catch (e) { }
-    
-    await fs.mkdir(targetHeaderDir);
-    
     await installNodeHeaders('0.25.2', null, targetHeaderDir);
     let canary = await fs.stat(path.join(targetHeaderDir, '.node-gyp', '0.25.2', 'common.gypi'));
     expect(canary).to.be.ok
 
-    await rimraf(targetHeaderDir);
+    await rmdir(targetHeaderDir);
   });
 
   it('check for installed headers for 0.27.2', async () => {
     let targetHeaderDir = path.join(__dirname, 'testheaders');
-
-    try {
-      if (await fs.stat(targetHeaderDir)) {
-        await rimraf(targetHeaderDir);
-      }
-    } catch (e) { }
-
-    await fs.mkdir(targetHeaderDir);
-    await fs.mkdir(path.join(targetHeaderDir, '.node-gyp'));
-    await fs.mkdir(path.join(targetHeaderDir, '.node-gyp', '0.27.2'));
+    await rmdir(targetHeaderDir);
+    await mkdir(targetHeaderDir);
+    await mkdir(path.join(targetHeaderDir, '.node-gyp'));
+    await mkdir(path.join(targetHeaderDir, '.node-gyp', '0.27.2'));
     const canary = path.join(targetHeaderDir, '.node-gyp', '0.27.2', 'common.gypi');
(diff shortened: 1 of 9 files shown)


cdaringe:
I closed temporarily because tests weren't passing.  However, even on master they are not passing.  Perhaps this is something you can assist with?

cdaringe:
testing has shown it's just not quite there yet :/

paulcbetts:
You don't have to close PRs if they're work in progress! Just add a TODO list in the description

cdaringe:
okee doke :).  can you talk to me about `prepublish: npm run compile`?  I've use npm lifecycle hooks often, but I'm thinking there may be some magic with publish?  When I'm trying to install my branch from git, nslog starts building before the `lib/` dir gets built.  what is supposed to trigger the lib build?  usually i've done `postinstall`

maxkorp:
when you run `npm install` inside a project, the prepublish script gets run after `postinstall`.

maxkorp:
Also, regarding your first list item, that's really an NPM issue that _still_ isn't resolved, even though people have been trying to get that handled for at least a year now. That said, maybe switching to postinstall might work.

cdaringe:
thanks @maxkorp.  It's pretty obnoxious. doing `npm i --save electron-rebuild` seems to do the prepublish before installing deps, whilst doing `npm i --save git+...electron-rebuild` tries to do it afterwards, thus the content isn't built!  arg!  i've temporarily added the lib/* to my branch, just so i can do some external testing.

paulcbetts:
@cdaringe Yeah, prepublish scripts don't run on `git://` dependencies. Do you know about `npm link`? It might make development easier.

Prepublish runs:

* When you run `npm install` in the root of your package that has a prepublish script
* Before you publish (obvs)

Prepublish does *not* run:

* When you run `npm install my-cool-package`

cdaringe:
thanks @paulcbetts, yea, I ended up l just `link`ing.  But I sure was going nuts before hand. :)  Fear not, I'll rebase the `lib/*` content out once it's all finished up.

cdaringe:
@paulcbetts @maxkorp, I think we're ready for review.  I figured we could run everything through eslint --fix or standard after all is said and done

test/main.js
```diff
@@ -62,38 +50,45 @@ describe('rebuildNativeModules', function() {
...
-          await rimraf(targetModulesDir);
-        }
-      } catch (e) { }
-
-      await installNodeHeaders(nativeModuleVersionToBuildAgainst, null, targetHeaderDir);
-      let canary = await fs.stat(path.join(targetHeaderDir, '.node-gyp', nativeModuleVersionToBuildAgainst, 'common.gypi'));
-      expect(canary).to.be.ok
-
-      // Copy our own node_modules folder to a fixture so we don't trash it
-      await cp(path.resolve(__dirname, '..', 'node_modules'), targetModulesDir);
```
cdaringe: let's not build our full node_modules folder, as it is pretty slow

test/main.js
```diff
@@ -62,38 +50,45 @@ describe('rebuildNativeModules', function() {
...
+          await rmdir(targetHeaderDir);
+          await mkdir(targetHeaderDir);
+          await rmdir(targetModulesDir);
+          await mkdir(targetModulesDir);
+          await installNodeHeaders(version, null, targetHeaderDir);
+          let canary = await fs.stat(path.join(targetHeaderDir, '.node-gyp', version, 'common.gypi'));
+          expect(canary).to.be.ok
+
+          rmdir('mock-pkg-1');
+          const mockPkg = pkgMock.install({
```
cdaringe: instead let's build a tiny fake pkg

test/main.js
paulcbetts: I actually don't mind slow tests, I like that the node_modules folder is a realistic test

cdaringe: > I actually don't mind slow tests,

I, respectfully, **very much mind** slow tests

> node_modules folder is a realistic test

i've proposed using a mocked module, which npm thinks is a _real_ module, that has actions post-build.  i then go in and assert that the build has completed, by testing that files exist that _would only exist if the build ran_.  I believe it to be very realistic, and much faster than building a set of modules from our src

paulcbetts: Nope! We're gonna stick with the realistic test, sorry! If you have a way to conditionally disable tests, I'd be into that

cdaringe: @paulcbetts, the test as it was previously written doesn't test any output or byproduct of the rebuild.  It just executes the rebuild, cleans up, and exits.  Do think there is a value in providing an explicit before and after condition, then asserting against it?  I would think that using the node_modules/ folder puts the vendor modules in control of the prebuild state and postbuild state, as opposed to an explicit before-and-after state (in my case, the absence then presence of the file).  further, what's the value of testing all modules at once vs. one explicit one?

Issue #26, opened 2015-10-12
Rebuild a single module


maxkorp:
I'd love for this to be able to rebuild a single node_module (in case it has been updated etc). I can add the functionality, but which flag would it go under? `-p` (for package)?

PR #29, opened 2015-10-13
Add flag for which node module to build

maxkorp:
I've added a flag `-w` to allow this to rebuild a single module, if you so choose. Otherwise, it still just rebuilds everything like before. See #26.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -18,6 +18,8 @@ const argv = require('yargs')
   .alias('a', 'arch')
   .describe('m', 'The path to the node_modules directory to rebuild')
   .alias('m', 'module-dir')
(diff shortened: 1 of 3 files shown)


paulcbetts:
Thanks @maxkorp!

maxkorp:
@paulcbetts Any chance of getting a build published now that this is in? Using it as a build dependency and I'd like to just keep it to a simple npm install.

paulcbetts:
@maxkorp Done, this is 1.0.2

maxkorp:
Awesome, thanks a ton dude. Just fyi doesn't look like the bump or tag got pushed to github.

paulcbetts:
@maxkorp Fixed

PR #34, opened 2015-11-05
Allow different npm commands via -c flag

maxkorp:
Added a flag for -c to allow you to pass different commands (i.e. install) to npm via the module. Still defaults to `rebuild`

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -96,7 +96,7 @@ shouldRebuildPromise
     if (!x) process.exit(0);
 
     return installNodeHeaders(argv.v, null, null, argv.a)
-      .then(() => rebuildNativeModules(argv.v, argv.m, argv.w, argv.c, null, argv.a))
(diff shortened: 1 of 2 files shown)


src/cli.js
```diff
@@ -88,7 +96,7 @@ shouldRebuildPromise
     if (!x) process.exit(0);
 
     return installNodeHeaders(argv.v, null, null, argv.a)
-      .then(() => rebuildNativeModules(argv.v, argv.m, argv.w, null, argv.a))
+      .then(() => rebuildNativeModules(argv.v, argv.m, argv.w, argv.c, null, argv.a))
```
paulcbetts: You can't bolt parameters into the middle like this, you break compatibility

maxkorp: That's fair, I'd added it there as it felt like it made the most sense, but I can move it to the end as well :+1:

PR #36, opened 2015-11-11
Update cli.js

MathijsvVelde:
[change cli.js](https://github.com/electronjs/electron-rebuild/issues/35)
Like this


zeke:
Best practice is to install `electron-prebuilt` as a devDependency so your project is portable and consistent across environments. This avoids the "works on my machine" problem.

PR #44, opened 2016-01-04
Workaround for node-pre-gyp's lack of electron runtime detection (and a little bugfixing)

addobandre:
[node-pre-gyp](https://www.npmjs.com/package/node-pre-gyp) is used by some popular projects (like [serialport](https://www.npmjs.com/package/serialport)) to provide precompiled binaries to most users. In electron's case it falls back to compilation and does so successfully.
Unfortunately, node-pre-gyp doesn't detect electron during runtime and tries to load native addons through the "normal" node path (...../node-v1234-os-arch.... instead of electron-v1234-os-arch).
Some pull requests have been submitted that fix the issue: mapbox/node-pre-gyp#177 and mapbox/node-pre-gyp#187 but they have been open for some months without any activity.

So I created a command-line switch (-p or --pre-gyp-fix) that copies the bindings' folder to the path current node-pre-gyp versions expect it to be in, when a fix is released in node-pre-gyp compatibility should not be broken and, since it's disabled by default it won't break anything to those not using it.

While developing this patch I noticed a couple of things
- electronPath in src/cli.js is defined without taking into account the path where electron-prebuilt resides, generating an ENOENT whenever getElectronModuleVersion() is called
- Calling toString(versionAsString) at the end of getElectronModuleVersion() turns the version string (it is returned as a string already) into a useless weird object

...so I fixed them in the process! :)

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -53,7 +53,7 @@ let nodeModuleVersion = null;
 if (!argv.n) {
   try {
     let pathDotText = path.join(argv.e, 'path.txt');
-    electronPath = path.join(argv.e, fs.readFileSync(pathDotText, 'utf8'));
+    electronPath = path.resolve(argv.e, fs.readFileSync(pathDotText, 'utf8'));
   } catch (e) {
     console.error("Couldn't find electron-prebuilt and no --node-module-version parameter set, always rebuilding");
   }
--- a/src/cli.js
+++ b/src/cli.js
@@ -87,16 +87,13 @@ if (!electronPath && !nodeModuleVersion) {
 }
 
 shouldRebuildPromise
-  .then(async (x) => {
-    var beforeRebuild;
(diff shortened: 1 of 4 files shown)


src/cli.js
```diff
@@ -50,7 +53,7 @@ let nodeModuleVersion = null;
 if (!argv.n) {
   try {
     let pathDotText = path.join(argv.e, 'path.txt');
-    electronPath = fs.readFileSync(pathDotText, 'utf8');
+    electronPath = path.join(argv.e, fs.readFileSync(pathDotText, 'utf8'));
```
felicienfrancois: You should use path.resolve instead of path.join, so it can handles both cases, when path.txt contains absolute or relative path.

paulcbetts:
Thanks!

PR #51, opened 2016-02-02
Document the node-pre-gyp workaround and make it easier to invoke

enlight:
Re-export the `preGypFixRun` function from `main.js` so it's easier to import from the outside.


paulcbetts:
@enlight Thanks!

paulcbetts:
This is 1.1.3

PR #67, opened 2016-05-13
Check for canary file in the `iojs` header folder aswell.

### LINKED ISSUE #66: Not working on Electron 1.0.1
After upgrading to `Electron` 1.0.1, I receive these errors:

```bash
ENOENT: no such file or directory, stat '/home/parroit/Desktop/repos/tunnels/node_modules/electron-rebuild/lib/headers/.node-gyp/1.0.1/common.gypi'
Error: ENOENT: no such file or directory, stat '/home/parroit/Desktop/repos/tunnels/node_modules/electron-rebuild/lib/headers/.node-gyp/1.0.1/common.gypi'
    at Error (native)
```

Looking at `/home/parroit/Desktop/repos/tunnels/node_modules/electron-rebuild/lib/headers/.node-gyp`, it only contains an `iojs-1.0.1` folder.

I'm not sure if you yet have to publish a version that support the new `Electron` release, or if there is some confusion between versions...

MarshallOfSound:
This PR fixes #66 

Basically `node-gyp` treats versions `1.x.y` to `3.x.y` as `iojs` so the major bump to `1.x.y` on electron caused node-gyp to download to a `iojs-{version}` folder instead of `{version}` folder.

This PR simply checks if either exists

/cc @parro-it @jostrander

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -2,20 +2,19 @@ import path from 'path';
 import _ from 'lodash';
 import childProcess from 'child_process';
 import spawn from './spawn';
-import promisify from './promisify';
 export { preGypFixRun } from './node-pre-gyp-fix';
 
-const fs = promisify(require('fs'));
+const fs = require('fs');
 
 const getHeadersRootDirForVersion = (version) => {
   return path.resolve(__dirname, 'headers');
 };
 
 const checkForInstalledHeaders = async function(nodeVersion, headersDir) {
-  const canary = path.join(headersDir, '.node-gyp', nodeVersion, 'common.gypi');
-  let stat = await fs.stat(canary);
+  const canaryNode = path.join(headersDir, '.node-gyp', nodeVersion, 'common.gypi');
+  const canaryIoJS = path.join(headersDir, '.node-gyp', `iojs-${nodeVersion}`, 'common.gypi');
 
-  if (!stat) throw new Error("Canary file 'common.gypi' doesn't exist");
+  if (!(fs.existsSync(canaryNode) || fs.existsSync(canaryIoJS))) throw new Error("Canary file 'common.gypi' doesn't exist");
   return true;
 };
 


parro-it:
Fixing the typo, I tested your fork and it work well!

MarshallOfSound:
@parro-it Lol, yeah, I fixed that typo in the generated dist files.  Not in the source files.  I am clever :eggplant: :laughing:

paulcbetts:
Hm, I like this approach but we should see what @zcbenz thinks before we try to hack our way around this

jostrander:
I'm a little curious as to how this is working as the iojs headers being downloaded differ from the node-v1.01 headers that are provided directly from the zcbenz bucket. If I'm not mistaken isn't the iojs headers actually `iojs-v1.0.1` where the `node-v1.0.1` headers are actually for node `5.1.0` as that's the version of node used by electron in 1.0.1? I did do an sha256sum on both files and they are different.

MarshallOfSound:
@jostrander I can't explain why it works :+1: But on my computer downloading those two files appear to be the same?

parro-it:
@jostrander I did not check the files for equality , but I can confirm that yesterday using the downloaded `iojs-v1.0.1` file (by manually renaming it) was not working, while after @MarshallOfSound path the rebuild is successfully working.

MarshallOfSound:
While doing the PR into `node-gyp` the server that served up the header files appeared to respond to URL's it didn't respond to before.  Perhaps someone on the electron team changed something server side :+1:

jostrander:
@MarshallOfSound pointed out to me that the only reason the checksums differ is the metadata of the .tar files. However once extracted a `diff` proved that both files were exactly the same.

paulcbetts:
Let's just get this out the door

paulcbetts:
Published as 1.1.4, thanks @MarshallOfSound

lipis:
Maybe a release here as well... https://github.com/electron/electron-rebuild/releases :)

paulcbetts:
@lipis Done

PR #89, opened 2016-07-27
Update the README with fixed examples, async/await, mention Node

ide:
- Fixes the examples so that the version numbers are no longer prefixed with `v`, which no longer works with Electron 1.3.1 on macOS at least. With `v` you get an error about common.gypi missing.
- Update one of the examples to use async/await and have fewer hardcoded values
- Change io.js references to Node.js

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 ## Electron-Rebuild
 
-This executable rebuilds native io.js modules against the version of io.js
-that your Electron project is using. This allows you to use native io.js
-modules in Electron apps without your system version of io.js matching exactly
+This executable rebuilds native Node.js modules against the version of Node.js
+that your Electron project is using. This allows you to use native Node.js
+modules in Electron apps without your system version of Node.js matching exactly
 (which is often not the case, and sometimes not even possible).
 
 ### How does it work?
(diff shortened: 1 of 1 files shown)


zeke:
I haven't tested that code but the doc changes look good to me.

ide:
I run almost the same code in my build script and it's working for what that's worth.

MarshallOfSound:
Node doesn't have native support for `async` and `await` yet does it?

Should we really be giving example syntax that won't work without boilerplating with a transpiler like babel?

paulcbetts:
I intentionally didn't use async/await so that people wouldn't think it was required

ide:
I think it depends on whether the intent is to provide copy-pasteable code or to communicate the usage clearly. @paulcbetts want me to change it back if you think it's going to cause issues?

zeke:
@ide are you okay with scrapping the refactored code from this PR and just leaving the readme fixes?

ide:
@zeke I reverted to the original code and added in the snippet to programmatically get the Electron version, if you think that nets out being helpful.

zeke:
Definitely helpful, if even just for this:

> With `v` you get an error about common.gypi missing.

PR #90, opened 2016-07-28
rename io.js to Node.js in readme

zeke:
io.js helped shepherd us into this beautiful future, but its time is up.

I see some other relics like:

```js
const checkForInstalledHeaders = async function(nodeVersion, headersDir) {
  const canaryNode = path.join(headersDir, '.node-gyp', nodeVersion, 'common.gypi');
  const canaryIoJS = path.join(headersDir, '.node-gyp', `iojs-${nodeVersion}`, 'common.gypi');

  if (!(fs.existsSync(canaryNode) || fs.existsSync(canaryIoJS))) throw new Error("Canary file 'common.gypi' doesn't exist");
  return true;
};
```

Can we do without that too, or is there some legacy reason to hang onto it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 ## Electron-Rebuild
 
-This executable rebuilds native io.js modules against the version of io.js
-that your Electron project is using. This allows you to use native io.js
-modules in Electron apps without your system version of io.js matching exactly
+This executable rebuilds native Node.js modules against the version of Node.js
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
I guess you figured out what that is for, but for others we need to because node-gyp expects versions `1.x` -> `3.x` to be iojs and expects the prefixes accordingly.  So even though we are using Electron when it hit `1.x` it expects to be called iojs

PR #92, opened 2016-08-04
Add new methods to find the electron-prebuilt directory

### LINKED ISSUE #91: Electron prebuilt is being renamed to electron
The `electron-prebuilt` module will be known as `electron` soon and as such this module needs to be updated to search for `electron` as well when determining the version.

The current workaround is to do `-e=node_modules/electron`

MarshallOfSound:
Fixes #91

This checks the `electron` node_modules directory and attempts to resolve the `electron` and `electron-prebuilt` packages as fallbacks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: node_js
+node_js:
+  - "6"
+  - "4"
+
+env:
+  - CXX=g++-4.8
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
+      - g++-4.8
--- a/src/cli.js
+++ b/src/cli.js
@@ -1,7 +1,8 @@
 #!/usr/bin/env node
 
 import {installNodeHeaders, rebuildNativeModules, shouldRebuildNativeModules} from './main.js';
-import { preGypFixRun } from './node-pre-gyp-fix.js'
+import {preGypFixRun} from './node-pre-gyp-fix.js'
+import {locateElectronPrebuilt} from './electron-locater';
 import path from 'path';
 import fs from 'fs';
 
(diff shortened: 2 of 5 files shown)


zeke:
Nice. @MarshallOfSound. Are you comfortable adding some tests for this?

Unfortunately, it looks like there are no CLI tests yet. 🙈

MarshallOfSound:
@zeke I'd rather put CLI tests in a separate PR as it would involve quite a bit of test bootstrapping and such that doesn't quite fit with this change.   Is that OK or do you want them in here anyway?

zeke:
The Electron team is holding off on a blog post about `electron` until 1.3.3 lands on Tuesday, August 9. That gives us a few days to smooth this out.

What about opening a separate PR for the CLI tests, and see how far we get by Tuesday. If the tests are ready by then, great. If not, we can roll this out as-is. Does that sound good?

MarshallOfSound:
Sounds good to me 👍

MarshallOfSound:
I came up with a neat way (IMO) to test just this functionality (that we can find electron if `electron` or `electron-prebuilt` is installed.  I'll push those tests up to this PR.

src/cli.js
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/env node
 
 import {installNodeHeaders, rebuildNativeModules, shouldRebuildNativeModules} from './main.js';
-import { preGypFixRun } from './node-pre-gyp-fix.js'
+import {preGypFixRun} from './node-pre-gyp-fix.js'
+import {locateElectronPrebuilt} from './electron-locater';
```
kevinsawicki: 🔥 `;`

test/electron-locater.js
```diff
@@ -0,0 +1,60 @@
+import _ from './support';
+import fs from 'fs';
+import npm from 'npm';
+import path from 'path';
+import promisify from '../lib/promisify.js';
```
kevinsawicki: This looks to be unused in this file.

test/electron-locater.js
```diff
@@ -0,0 +1,60 @@
+import _ from './support';
+import fs from 'fs';
+import npm from 'npm';
```
kevinsawicki: This looks like it might be unused as well.

.travis.yml
```diff
@@ -0,0 +1,13 @@
+language: node_js
```
kevinsawicki: Thanks for adding this file 👍 🔧

MarshallOfSound:
@zeke Ping RE this 👍

zeke:
Landed in 1.2.0

PR #94, opened 2016-08-05
Add a Changelog

zeke:
This is an attempt to add a changelog to the project. The goal is to have something that's easier for humans to read than a scroll-fest through the commit history.

The publishing process has some manual steps here, but if folks feel like this changelog is worth having, we could pretty easily create a release script that would turn the whole thing into a one-liner. Something like this:

```sh
npm run release -- minor
```

_Note that the 1.2.0 version here is optimistic; current published version is still 1.1.4_

cc @kevinsawicki @jlord @paulcbetts

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: node_js
+node_js:
+  - "6"
+  - "4"
+
+env:
+  - CXX=g++-4.8
+addons:
+  apt:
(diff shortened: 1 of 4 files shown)


zeke:
No strong feelings here, apparently. Closing and we can revisit if we ever want a changelog.

PR #98, opened 2016-08-12
Add --ignore-optdeps and --ignore-devdeps CLI params

### LINKED ISSUE #61: Verbose output option
Depending on what packages it has to deal with, `electron-rebuild` sometimes takes a good long while; @paulcbetts would you be open adding a 'verbose' CLI option that shows that it's doing?

### LINKED ISSUE #96: Electron rebuild shouldn't rebuild devDependencies
Currently (as far as I can tell) `electron-rebuild` rebuilds `devDependencies`.

We can change this by passing in a list of package names to `npm rebuild` when we spawn it in `rebuildNativeModules`.

I've almost got a working branch with this change however I was wondering how people want to handle `optionalDependencies`

For instance I currently have an optional dependency that is technically a dev dependency but it's optional so it can fail.

TLDR: I want to change `electron-rebuild` so it doesn't rebuild devDeps, what should it do for optDeps

/cc @paulcbetts @zeke ?

MarshallOfSound:
Fixes #96 
Fixes #61 

This also sneaks in a `--log` arg which automatically pipes the output of the rebuild to `process.stdout`

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -20,7 +20,7 @@ const yargs = require('yargs')
   .alias('a', 'arch')
   .describe('m', 'The path to the node_modules directory to rebuild')
   .alias('m', 'module-dir')
-  .describe('w', 'A specific module to build')
+  .describe('w', 'A specific module to build, or comma separated list of modules')
   .alias('w', 'which-module')
   .describe('e', 'The path to electron-prebuilt')
   .alias('e', 'electron-prebuilt-dir')
@@ -30,6 +30,12 @@ const yargs = require('yargs')
   .alias('c', 'command')
   .describe('d', 'Custom header tarball URL')
   .alias('d', 'dist-url')
+  .describe('id', 'Ignore devDependencies')
+  .alias('id', 'ignore-devdeps')
+  .describe('io', 'Ignore optionalDependencies')
+  .alias('io', 'ignore-optdeps')
+  .describe('l', 'Log the rebuild process')
+  .alias('l', 'log')
   .epilog('Copyright 2015');
 
 const argv = yargs.argv;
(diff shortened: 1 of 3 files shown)


MarshallOfSound:
Pinging @paulcbetts @zeke @mnquintana

src/main.js
```diff
@@ -116,7 +116,20 @@ export async function rebuildNativeModules(nodeVersion, nodeModulesPath, whichMo
   ];
 
   if (whichModule) {
-    args.push(whichModule);
+    args = args.concat(whichModule.split(','));
+  }
+
+  if (ignoreDevDeps || ignoreOptDeps) {
+    console.log(nodeModulesPath)
```
zeke: Did you leave this logging statement in here intentionally?

zeke:
I like the idea, but am not familiar enough with the codebase yet to make a judgement of the implementation. Are you able to add tests for it?

>  I was wondering how people want to handle optionalDependencies

I have never used optional dependencies, and I very rarely see them, so I suspect this is not a desperately needed feature.

MarshallOfSound:
So this has stalled miserably.  I just hit another instance of wanting something like this.  Can I possibly get a go / no go from the powers at be 👍 

@electron/maintainers

src/cli.js
```diff
@@ -111,7 +117,7 @@ shouldRebuildPromise
   })
   .then((x, beforeRebuild) => {
     return installNodeHeaders(argv.v, argv.d, null, argv.a)
-      .then(() => rebuildNativeModules(argv.v, argv.m, argv.w, null, argv.a, argv.c))
+      .then(() => rebuildNativeModules(argv.v, argv.m, argv.w, null, argv.a, argv.c, argv.id, argv.io, argv.log))
```
enlight: These one/two-letter variable names drive me nuts, can you use the aliases here instead? As in `argv.ignoreDevdeps`, `argv.ignoreOptdeps`. We can clean up the rest later.

MarshallOfSound: The CLI tool didn't appear to work that way off the top of my head, the alias mapped to the shorthand 😢

src/main.js
MarshallOfSound: Woops, didn't even see this comment either 😆

src/cli.js
enlight: `yargs` will create a prop on `yargs.argv` for the key, the alias (verbatim), and the alias converted to camel-case, at least the versions I've used have done just that.

MarshallOfSound: I shall play around and see what can be done 👍

PR #114, opened 2016-10-09
Add verbose mode

Paxa:


src/logger.js
```diff
@@ -0,0 +1,26 @@
+import colors from 'colors/safe';
+
+var colorList = ['blue', 'yellow', 'magenta', 'cyan'];
+
+var enabled = !!process.env.DEBUG;
```
MarshallOfSound: This should be `const`

src/logger.js
```diff
@@ -0,0 +1,26 @@
+import colors from 'colors/safe';
+
+var colorList = ['blue', 'yellow', 'magenta', 'cyan'];
```
MarshallOfSound: This should be `const`

src/logger.js
```diff
@@ -0,0 +1,26 @@
...
+
+var enabled = !!process.env.DEBUG;
+
+export default function logger (command, ...messages) {
+  if (enabled) {
+    var color = colorList[command.length % colorList.length];
```
MarshallOfSound: I don't get the point of these color calls.  Can you explain?

To me it looks like we are almost randomly choosing a color for the given console message

src/logger.js
```diff
@@ -0,0 +1,26 @@
...
+    console.log.apply(console, [colors[color](command)].concat(messages));
+  }
+}
+
+logger.print = function print (...data) {
```
MarshallOfSound: Attaching methods as a property of a function doesn't make a whole lot of sense.  It would probably make more sense to export these methods as named constants.

I.e.

```js
export const print = function (...data) {
  ...
}
```

src/main.js
```diff
@@ -124,6 +128,7 @@ export async function rebuildNativeModules(nodeVersion, nodeModulesPath, whichMo
     `--target=${nodeVersion}`,
     `--arch=${arch || process.arch}`
   );
+  if (process.env.DEBUG) args.push('--verbose');
```
MarshallOfSound: I might be wrong here, but the `DEBUG` variable is normally set to a glob string that matches package names that you want to debug.  So just because `DEBUG` is not undefined does not mean the user wants to debug this specific module.

test_project/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "version": "0.1.0",
+  "repository": "https://github.com/electron/electron-rebuild",
+  "license": "MIT",
+  "scripts": {
+    "start": "electron .",
+    "rebuild": "electron-rebuild -f -p --verbose"
+  },
+  "devDependencies": {
+    "electron": "^1.4.3",
+    "electron-rebuild": "github:electron/electron-rebuild"
```
MarshallOfSound: This should not work and should be be reccomended in the source code.  Because our output is `babelified` just cloning from github will not result in a workable module.

test_project/index.js
```diff
@@ -0,0 +1,17 @@
+const electron = require('electron');
```
MarshallOfSound: What is the overall purpose of the `test_project`. Is it used to run tests somewhere or am I missing something?

src/logger.js
Paxa: It will calculate color base on first argument length, so same function will have same color

MarshallOfSound: @Paxa Yes, but what is the point of changing the color for different functions.  In fact, what is the point in having colors at all.  Personally I find colors annoying in CLI tools and having seemingly random colors assigned to random logs would confuse more people than it helps.

For instance "yellow" would be interpreted as a warning., blue as info. Etc.  The easiest thing to do is to leave colors out completely

Paxa: IMO with colors is more readable. Colors are kinda random but for same command there will be same color

screen shot 2016-10-10 at 11 21 40 am
vs
screen shot 2016-10-10 at 11 22 46 am

we can exclude yellow
screen shot 2016-10-10 at 11 26 00 am

MarshallOfSound: Why not just consistently use the same colour?

Paxa: Ok, what colour you prefer?

MarshallOfSound: I don't mind, as long as it is readable on dark backgrounds (normal terminal background is darkish) and it remains consistent 👍

src/logger.js
```diff
@@ -0,0 +1,24 @@
...
+const colorList = ['blue', 'magenta', 'cyan'];
+let enabled = false;
+
+export default function logger (command, ...messages) {
+  if (enabled) {
+    if (messages.length == 0) {
+      console.log(command);
+    } else {
+      var color = colorList[command.length % colorList.length];
+      console.log.apply(console, [colors[color](command)].concat(messages));
```
MarshallOfSound: Magical ES6 features allow us to change this to be

```js
console.log(colors[color](command), ...messages);
```

src/logger.js
Paxa: Change it to cyan

src/logger.js
Paxa: Done

Paxa:
Synced with latest master

Paxa:
Any thoughts? Anything else to change?

MarshallOfSound:
Due to the rewrite of `electron-rebuild` I don't believe this PR can still be applied.  Thanks for the original effort though 👍

Paxa:
Would you like to have it for new code base?

MarshallOfSound:
@Paxa The new code base throws stderr/out automatically when it fails 👍

PR #115, opened 2016-10-10
Create testing application

Paxa:
For:
- testing some edge cases
- example
- integration test


MarshallOfSound:
Due to the rewrite of `electron-rebuild` I don't believe this PR can still be applied.  Thanks for the original effort though 👍

PR #116, opened 2016-10-12
Always update the binary on rebuild

### LINKED ISSUE #59: Best practice to support electron-rebuild
Hi,
I'm a maintainer of [`node-serialport`](https://github.com/voodootikigod/node-serialport) and a bunch of my users use electron and `electron-rebuild`. We hadn't been supporting it as well as I'd like and with our latest beta we've upgraded `nan` and cleaned up our build processes. This has broth improved electron support as we work with the latest versions and made it more annoying.

According to [this issue](https://github.com/voodootikigod/node-serialport/issues/538#issuecomment-207852190) we can now run `electron-rebuild` but we must first remove our binary file manually. It currently resides in `build/Release/serialport.node` It will be there if you've built it or downloaded it via `pre-gyp` during install. That seems annoying. I'm a bit new to the electron ecosystem so I wanted to check with you to see if we could do something to make this easier, or provide better instructions to get it going. 

Thanks!

reconbot:
This either downloads or rebuilds the binary module during rebuild.

Closes #59

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -122,7 +122,8 @@ export async function rebuildNativeModules(nodeVersion, nodeModulesPath, whichMo
   args.push(
     '--runtime=electron',
     `--target=${nodeVersion}`,
-    `--arch=${arch || process.arch}`
+    `--arch=${arch || process.arch}`,
+    `--update-binary`
   );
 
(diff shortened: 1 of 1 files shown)


paulcbetts:
Seems legit, thanks for the PR @reconbot

PR #120, opened 2016-10-22
WIP Add a electron-packager afterCopy hook to the exports

### LINKED ISSUE #99: Export a function compatible with electron-packager afterCopy hook
I recently PR'ed into `electron-packager` to add a hook to their API that will run custom functions once it has successfully copied your app files to a temp folder.

https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#aftercopy

I want to make a function that will receive the params from that hook call and rebuild the node modules in that temporary folder so that the node modules in your actual directory remain untouched

MarshallOfSound:
This adds a `createPackagerCopyHook` function to the exports of `electron-rebuild`.

Basic usage is like so

``` js
packager({
  ...
  afterCopy: [createPackagerCopyHook()]
  ...
})
```

You can add custom options like so

``` js
createPackagerCopyHook(distUrl, ignoreDev, ignoreOpt, preGypFix, verbose)
```

This makes rebuilding modules for distribution so much easier as you don't have to worry about your node_modules folder being built for a different arch after you run your packager.

Fixes #99

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -2,7 +2,10 @@ import path from 'path';
 import _ from 'lodash';
 import childProcess from 'child_process';
 import spawn from './spawn';
-export { preGypFixRun } from './node-pre-gyp-fix';
+import { preGypFixRun } from './node-pre-gyp-fix';
+import { locateElectronPrebuilt } from './electron-locater';
+
+export { preGypFixRun };
 
 const fs = require('fs');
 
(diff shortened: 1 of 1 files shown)


paulcbetts:
@MarshallOfSound Do we still want to do this?

MarshallOfSound:
Nope, the newly exported function is already basically in the required form for an afterCopy hook.  This has been superceded :+1:

PR #123, opened 2016-11-14
Export more methods and make an index.js file

paulcbetts:
This PR exports the `locateElectronPrebuilt` method as well as makes it a bit easier to decide what gets exported and what doesn't -cc @zeke


zeke:
👋 Any particular reason you CC'ed me?

PR #138, opened 2017-01-05
Handle the case where the module specifies an output path

### LINKED ISSUE #137: Placeholders in the paths
Hello,

Latest version (1.5.5) stores build artifacts in the directories with placeholders in names. Then electron runtime can't find them and nothing works.

image

image

I am on node 6.7, running `electron-rebuild` from a script with the following params:

image

MarshallOfSound:
* Replaces the standard node-gyp bindings template strings

Fixes #137

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: node_js
 node_js:
 - '6'
-- '4'
 
 sudo: false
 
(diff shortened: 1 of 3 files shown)


MarshallOfSound:
Looks like the `spawn-rx` update broke node v4 support.  @paulcbetts Do we want to continue supporting v4 or force the move to the better and brighter v6 👍

paulcbetts:
I'm okay with :fire:ing node v4

malept:
I suggest mentioning it in the README, too.

PR #159, opened 2017-03-28
Refactor to TypeScript

paulcbetts:
This PR refactors electron-rebuild to be in TypeScript, and cleans up rebuild a bit. There should be no functional changes, nor should this be a breaking change

## TODO:

- [ ] Figure out what to do about default exports
- [x] Make sure the tests pass


src/main.ts
```diff
@@ -1,4 +1,4 @@
-export { default, rebuildNativeModules } from './rebuild';
+export { rebuild, rebuildNativeModules } from './rebuild';
```
bendemboski: @paulcbetts @MarshallOfSound this is a breaking change -- shouldn't it have been released as a major version bump? It broke `electron-forge` -- I think [this](https://github.com/electron-userland/electron-forge/blob/master/src/util/rebuild.js#L1) line needs to be `import { rebuild } from 'electron-rebuild'`.

MarshallOfSound: Bugger, will fix and patch ASAP

MarshallOfSound: @bendemboski Released a patch fix now to export on `default` and `rebuild` and manually define `__esmodule` so that babel picks up the default export

PR #164, opened 2017-04-15
Don't rebuild modules that have been prebuilt with prebuild

### LINKED ISSUE #162: Support modules with bundled prebuilds for electron
All new native modules I've been writing, use a new technique where I bundle all prebuilds inside the npm tarball and the pick the right one at runtime. A huge benefit of this is that using these modules with electron *just works*, no compilation needed.

However when people use electron-rebuild, it forces the prebuilded modules to rebuild by running `node-gyp rebuild` instead of running the install script if present in package.json. Ironically this breaks some of my modules that support electron out of the box cause it tries to recompile them without running a preinstall script specificed by the install script.

Anyway to fix this?

As an example try my [sodium-native](https://github.com/mafintosh/sodium-native) module. It is a native module that works across node/electron without needing any recompilation that breaks when electron-rebuild tries to recompile it, cause it isn't running the install script instead node-gyp like npm does.

MarshallOfSound:
Fixes #162 

/cc @mafintosh

### DIFF
--- a/test/fixture/native-app1/package.json
+++ b/test/fixture/native-app1/package.json
@@ -19,7 +19,8 @@
   "dependencies": {
     "ref": "1.3.3",
     "benchr": "3.2.0",
-    "@newrelic/native-metrics": "1.0.0"
+    "@newrelic/native-metrics": "1.0.0",
+    "sodium-native": "1.8.0"
   },
   "optionalDependencies": {
     "zipfile": "0.5.11"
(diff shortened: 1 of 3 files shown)


test/fixture/native-app1/package.json
```diff
@@ -19,7 +19,8 @@
   "dependencies": {
     "ref": "1.3.3",
     "benchr": "3.2.0",
-    "@newrelic/native-metrics": "1.0.0"
+    "@newrelic/native-metrics": "1.0.0",
+    "sodium-native": "1.8.0"
```
paulcbetts: What do we do with sodium-native?

PR #170, opened 2017-05-09
add node-pre-gyp support

b-ono:
This PR modifies electron-rebuild to use node-pre-gyp if the binary section is present. It uses the --fallback-to-build option to build the dependency if no prebuild binary is available.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "fs-promise": "^2.0.2",
     "node-abi": "^2.0.0",
     "node-gyp": "^3.6.0",
+    "node-pre-gyp": "^0.6.34",
     "ora": "^1.2.0",
     "rimraf": "^2.6.1",
     "spawn-rx": "^2.0.10",
--- a/src/rebuild.ts
+++ b/src/rebuild.ts
@@ -175,7 +175,9 @@ class Rebuilder {
     });
 
     d('built:', moduleName);
-    await fs.mkdirs(path.dirname(metaPath));
+    if (!(await fs.exists(metaPath))) {
+      await fs.mkdirs(path.dirname(metaPath));
+    }
     await fs.writeFile(metaPath, metaData);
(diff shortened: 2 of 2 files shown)


MarshallOfSound:
Can you explain exactly what this achieves, as far as I am aware `node-pre-gyp` is 100% incompatible with Electron binaries and this `fallback-to-build` is semantically equivalent to our existing `node-gyp` logic?

b-ono:
node-pre-gyp can build the packages for electronif the runtime and disturl and so on is specified correctly. It can also download the prebuild packages from the specified host. 

`fallback-to-build` should be equivalent to the existing `node-gyp` logic.

MarshallOfSound:
As far as I am aware `node-pre-gyp --fallback-to-build` is just a complicated wrapper around `node-gyp` and this module actually exists to fix numerous issues with using `node-pre-gyp` so I'd like to avoid using it.

If this PR actually fixes a bug please raise that as an issue and we can look at solving it without bring in the `pre-gyp` dependency

roytan883:
@MarshallOfSound `node-pre-gyp` is very useful when the module already has binary pre-build. for example, use electron with `realm-js` which has:
```
  "binary": {
    "module_name": "realm",
    "module_path": "./compiled/{node_abi}_{platform}_{arch}/",
    "host": "https://static.realm.io",
    "remote_path": "/node-pre-gyp/{version}/"
  }
```
so `node-pre-gyp` will download it from 
`https://static.realm.io/node-pre-gyp/2.29.2/realm-v2.29.2-electron-v4.2-win32-x64.tar.gz`.
This `tar.gz` only include a  single file `realm.node` already pre-build.

please consider add `node-pre-gyp` support. Because `electron-forge` depend on `electron-rebuild` but `electron-rebuild` not support `node-pre-gyp`.

PR #178, opened 2017-05-29
fix: do not ignore input modules provided via -w

jkuri:
in my case `node-sass` was recompiling over and over again (2-3mins) and I couldn't provide option to build only modules I want (`-w`) or to ignore others. 
this fixes the behaviour of `-w` flag. 

closes issue: https://github.com/electron/electron-rebuild/issues/151

### DIFF
--- a/src/rebuild.ts
+++ b/src/rebuild.ts
@@ -26,7 +26,7 @@ const locateNodeGyp = async () => {
 class Rebuilder {
   ABI: string;
   nodeGypPath: string;
-  prodDeps: Set<string>;
+  prodDeps: { [key: string]: boolean };
   rebuilds: (() => Promise<void>)[];
 
   constructor(
@@ -38,9 +38,10 @@ class Rebuilder {
       public forceRebuild = false,
       public headerURL = 'https://atom.io/download/electron',
       public types = ['prod', 'optional'],
-      public mode = defaultMode) {
+      public mode = defaultMode,
+      public onlyModules: string[] = []) {
     this.ABI = nodeAbi.getAbi(electronVersion, 'electron');
-    this.prodDeps = extraModules.reduce((acc, x) => acc.add(x), new Set());
+    this.prodDeps = {};
     this.rebuilds = [];
   }
 
@@ -53,7 +54,6 @@ class Rebuilder {
     this.lifecycle.emit('start');
 
     const rootPackageJson = await readPackageJson(this.buildPath);
-    const markWaiters: Promise<void>[] = [];
     const depKeys = [];
 
     if (this.types.indexOf('prod') !== -1) {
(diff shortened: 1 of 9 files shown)


MarshallOfSound:
I believe the docs need cleaning up around what `-w` actually does as in my opinion `-w` is working 100% as expected, even in the code base the variable is `extraModules` not `onlyModules`.

I like this feature but I think the `-w` flag needs to remain, this should be under a new option.  Let's say `--only`.

src/rebuild.ts
```diff
@@ -53,22 +53,30 @@ class Rebuilder {
     this.lifecycle.emit('start');
 
     const rootPackageJson = await readPackageJson(this.buildPath);
-    const markWaiters: Promise<void>[] = [];
-    const depKeys = [];
+    const markWaiters: Promise<void>[] = this.types.reduce((acc: string[], curr: string) => {
```
MarshallOfSound: I personally don't like this `reduce` change, sure it's nice and reduce is a pretty cool ES6 feature but in this case I think the 3 consecutive if statements where much easier to understand than a reduce call

jkuri: hi @MarshallOfSound. thanks for a quick response. 
here I tried to avoid additional variable which is not needed in the future, but can revert that ofc.

jkuri:
@MarshallOfSound PR updated. please re-review.

MarshallOfSound:
Looking good, can we get a test added for this 👍

jkuri:
added two tests for checking new option. travis is still failing but actually the very first test which I didn't change at all:
```sh
locateElectronPrebuilt
    1) should return null when electron is not installed
```
locally everything seems to be good (checked on Linux and macOS). 
maybe any idea what should cause this initial test to fail?

MarshallOfSound:
@jkuri It looks like travis isn't cleaning up their build machines and `electron's` build is also on that machine at exactly the correct search path 😢

jkuri:
@MarshallOfSound I was pulling my hear trying to fix this test, then checked my fork builds and it passed every time, check https://travis-ci.org/jkuri/electron-rebuild/builds/237074694. 

can you please try clearing cache on travis (see below screenshot) and manually restarting the build?

travis-cache

btw, while figuring this out I "merged" previous PR of mine with this one, so if this PR will be accepted, https://github.com/electron/electron-rebuild/pull/177 can be closed. thanks.

MarshallOfSound:
@jkuri Looks like you got stuck on the bad end of a merge conflict.  If you merge in the latest master the travis builds will start passing and then I can get this in 👍

jkuri:
closing in favor of https://github.com/electron/electron-rebuild/pull/190

PR #181, opened 2017-06-08
Replace fs-promise with fs-extra

albinekb:
fs-promise is deprecated 👍 

>npm WARN deprecated fs-promise@2.0.3: Use mz or fs-extra^3.0 with Promise Support

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,6 @@ branches:
   - master
   - /^v\d+\.\d+\.\d+/
 
+before_script:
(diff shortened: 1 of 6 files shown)


albinekb:
```$ mocha --compilers ts:ts-node/register ./test/*.ts


  locateElectronPrebuilt
    ✓ should return null when electron is not installed
    with electron-prebuilt installed
      ✓ should return a valid path
    with electron installed
      ✓ should return a valid path

  read-package-json
    ✓ should find a package.json file from the given directory (53ms)
```

passes locally, seems that travis has electron installed by default?

MarshallOfSound:
Thanks 👍

PR #183, opened 2017-06-14
Support for cnpm and use electron as devDependence instead of electro…

wenshiqi0:
I am using `cnpm` instead of `npm` or `yarn` to install my node modules, and I find out that electron-rebuild does not support for `cnpm`. `cnpm` use link to handle the dependencies, as I can see electron-rebuild may make circles while scanning the dependencies tree and will rebuild a module twice and this will make node-gyp throw errors.

### DIFF
--- a/src/rebuild.ts
+++ b/src/rebuild.ts
@@ -28,6 +28,8 @@ class Rebuilder {
   nodeGypPath: string;
   prodDeps: Set<string>;
   rebuilds: (() => Promise<void>)[];
+  realModulePaths: Array<string>;
+  realNodeModulesPaths: Array<string>;
 
   constructor(
       public lifecycle: EventEmitter,
@@ -42,6 +44,8 @@ class Rebuilder {
     this.ABI = nodeAbi.getAbi(electronVersion, 'electron');
     this.prodDeps = extraModules.reduce((acc, x) => acc.add(x), new Set());
     this.rebuilds = [];
+    this.realModulePaths = [];
+    this.realNodeModulesPaths = [];
   }
 
   async rebuild() {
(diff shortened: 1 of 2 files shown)


fengmk2:
pnpm is using link too https://github.com/pnpm/pnpm

src/rebuild.ts
```diff
@@ -42,6 +44,8 @@ class Rebuilder {
     this.ABI = nodeAbi.getAbi(electronVersion, 'electron');
     this.prodDeps = extraModules.reduce((acc, x) => acc.add(x), new Set());
     this.rebuilds = [];
+    this.realModulePaths = [];
```
MarshallOfSound: For performance reasons these two path arrays should be sets

fengmk2:
@zcbenz we find the root cause finally.

src/rebuild.ts
```diff
@@ -184,19 +188,40 @@ class Rebuilder {
   }
 
   rebuildAllModulesIn(nodeModulesPath: string, prefix = '') {
-    d('scanning:', nodeModulesPath);
+    // While we use `cnpm`, it will make a circle scanning the dep tree.
+    // We also need to ensure that the `node_modules` which we are scanning has never came before.
+    const realNodeModulesPath = fs.realpathSync(nodeModulesPath);
+    if (this.realNodeModulesPaths.indexOf(realNodeModulesPath) > -1) {
+      return;
+    } else {
```
MarshallOfSound: No need for this else statement as their is a `return` in the main if condition

src/rebuild.ts
```diff
@@ -184,19 +188,40 @@ class Rebuilder {
...
+    for (const modulePath of fs.readdirSync(realNodeModulesPath)) {
+      // If we use `cnpm` to install mudules, it will be rebuilded failed,
+      // because of some same modules rebuilded twice or more.
+      // Need to ensure one dep only be rebuilded once.
+      const finalPath = path.resolve(nodeModulesPath, modulePath);
+      const realPath = fs.realpathSync(finalPath);
+
+      if (this.realModulePaths.indexOf(realPath) > -1) {
+        continue;
+      } else {
```
MarshallOfSound: Again no need for the else block as their is a `continue` statement in the main if block

src/rebuild.ts
```diff
@@ -42,6 +44,8 @@ class Rebuilder {
     this.ABI = nodeAbi.getAbi(electronVersion, 'electron');
     this.prodDeps = extraModules.reduce((acc, x) => acc.add(x), new Set());
     this.rebuilds = [];
+    this.realModulePaths = [];
+    this.realNodeModulesPaths = [];
```
fengmk2: use `Map` instead `Array` will be more faster.

MarshallOfSound: See my comment below about using sets, all we need is a concept of "has been used before", we don't need to store a value with the path so maps will be less efficient than using a raw set

MarshallOfSound:
Also if you pull in latest master the travis builds should go green 👍

wenshiqi0:
Need some Windows test.

wenshiqi0:
Test passed, but I find some more problems. Did electron-rebuild will handle `fsevents` rebuilding for Windows? Event I am in Windows, the `fsevents` module may still be rebuilded, and this make errors. @MarshallOfSound

PR #190, opened 2017-06-27
Add --only option to build whitelisted modules

### LINKED ISSUE #151: -w gets ignored
I am trying to rebuild the package `vlc` like so

```
./node_modules/.bin/electron-rebuild -v 1.4.15 -f -w vlc
```

but instead of rebuilding `vlc` another package, `sqlite3`, gets rebuild

> Building module: sqlite3, Completed: [num]

and then it finishes. when I provide no package with `-w` I see the same output (same package being rebuilt)

mojodna:
Adapted from changes originally in #178 (can be considered a replacement for that, with the additional benefit of handling nested module names).

Fixes #151

/cc @jkuri

### DIFF
--- a/src/rebuild.ts
+++ b/src/rebuild.ts
@@ -15,6 +15,7 @@ export interface RebuildOptions {
   electronVersion: string;
   arch?: string;
   extraModules?: string[];
+  onlyModules?: string[];
   force?: boolean;
   headerURL?: string;
   types?: ModuleType[];
@@ -55,6 +56,7 @@ class Rebuilder {
   public electronVersion: string;
   public arch: string;
   public extraModules: string[];
+  public onlyModules: string[];
   public force: boolean;
   public headerURL: string;
   public types: ModuleType[];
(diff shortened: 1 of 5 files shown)


sedwards2009:
@mojodna  @MarshallOfSound  The failure on travis doesn't appear to directly related to the health of this PR. Can someone have a closer look at this and/or give travis a kick plz?

sedwards2009:
@MarshallOfSound  thanks a lot man.

src/rebuild.ts
```diff
@@ -315,7 +315,7 @@ function doRebuild(options: any, ...args: any[]) {
   if (typeof options === 'object') {
     return rebuildWithOptions(options as RebuildOptions);
   }
-  console.warn('You are using the deprecated electron-rebuild API, please switch to using the options object instead');
+  console.warn('You are using the depreceated electron-rebuild API, please switch to using the options object instead');
```
mojodna: This actually was the typo fix: https://en.wikipedia.org/wiki/Deprecation

PR #201, opened 2017-09-09
docs: include Node version requirement

dland512:
specifies that node v6.0.0 or greater is required

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ and then
 npm run rebuild
 ```
 
-### What versions of node does it work with?
+### What versions of Node does it work with?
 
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
Isn't this in the package.json file and enforced by package managers? 😕

malept:
It's not enforced by NPM but it does emit a warning.

README.md
```diff
@@ -38,6 +38,11 @@ and then
 npm run rebuild
 ```
 
+### What versions of node does it work with?
```
malept: Might as well capitalize "Node"

PR #231, opened 2018-03-11
Add docs on how electron-rebuild works with prebuild-install

ralphtheninja:
I think some docs is in order. The support for `prebuild-install` is completely hidden in the code.

Closes https://github.com/electron/electron-rebuild/issues/229

cc @juliangruber @vweevers

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,10 @@ packager({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


ralphtheninja:
Hello?

zeke:
👋

PR #242, opened 2018-06-24
feat: add a caching option

MarshallOfSound:
* Generates a deterministic hash of each directory being rebuilt
  * hash is based on options passed to electron-rebuild and contents of
    directory
* Checks if a cache entry exists for that hash
* If not, builds like normal
  * Then stores the entire state of that module directory as a gzipped
    blob in the users cache directory
* If yes, pulls the module directory state and applies it directly to
  that module

 Please note that this is kinda experimental, I would NOT reccomend using
 it on CI, but for local dev where rebuilding modules constantly can get
 very annoying I would 100% reccomend using it.

This is delibrately being left undocumented so it can stablize a bit or
potentially be removed at any point.

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -1,7 +1,6 @@
 import * as crypto from 'crypto';
 import * as fs from 'fs-extra';
 import * as path from 'path';
-
 import * as zlib from 'zlib';
 
 class Snap {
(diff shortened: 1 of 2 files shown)


cloudsere:
Hi, thanks for your wonderful work!  However, I found that `useCache` and `cachePath` are not accepted by the cli as options...? Did I miss something?

malept:
This is not provided via the CLI, only the API. Presumably, because it's experimental.

PR #244, opened 2018-07-23
Add feature to allow users to specify proxy

enderian:
Because I am behind a corporate proxy, I could not use electron-rebuild without modifying the node-gyp invocation arguments to include a proxy.

So I implemented an argument for electron-rebuild to specify a proxy to be used by node-gyp.
The arguments is `-x` or `--proxy`, and it's as simple as calling:

> `electron-rebuild --proxy http://username:password@proxy.example.com:8080`

When executing gyp, electron-rebuild will push the --proxy argument to it.
Documentation for the flag can be found here: [Gyp Command Options](https://github.com/nodejs/node-gyp#command-options)


package.json
```diff
@@ -34,6 +34,8 @@
     "colors": "^1.2.0",
     "debug": "^2.6.3",
     "detect-libc": "^1.0.3",
+    "electron": "^2.0.5",
```
MarshallOfSound: Please revert this package.json change

PR #259, opened 2018-11-06
update node-abi 's version to 2.5.0

boneVidy:
when i update my electron version to 3.6.0, and excute `electron-rebuild` ,
I am struggling with the error:
`Could not detect abi for version 3.0.6 and runtime electron.  Updating "node-abi" might help solve this issue if it is a new release of electron`

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "debug": "^2.6.3",
     "detect-libc": "^1.0.3",
     "fs-extra": "^3.0.1",
-    "node-abi": "^2.0.0",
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
`^2.5.0` is within range of `^2.0.0`.  All you need to do is run `npm upgrade electron-rebuild` or `npm upgrade node-abi --depth=4` or `yarn upgrade electron-rebuild` (depends on what npm / yarn you are using)

boneVidy:
thx

PR #266, opened 2018-12-17
feat: add --prebuild-tag-prefix option for prebuild-install

HipsterBrown:
* passes through --prebuilt-tag-prefix option to prebuild-install call to get
scoped modules from github

This will help [unblock users of `node-serialport`](https://github.com/node-serialport/node-serialport/issues/1672#issuecomment-448016260) to use the latest prebuilt binaries and any other package that uses custom tag prefixes with electron-rebuild.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ Options:
   -o, --only                   Only build specified module, or comma separated
                                list of modules. All others are ignored.
   -b, --debug                  Build debug version of modules
+  -tp, --prebuild-tag-prefix   github tag prefix, default is "v", passed to prebuild-install
 
 Copyright 2016
 ```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -65,6 +65,7 @@ Options:
   -o, --only                   Only build specified module, or comma separated
                                list of modules. All others are ignored.
   -b, --debug                  Build debug version of modules
+  -tp, --tag-prefix            github tag prefix, default is "v"
```
MarshallOfSound: Can this be `--prebuild-tag-prefix` to be clear this is just for prebuild stuff

HipsterBrown: Sure thing. Will update.

Issue #271, opened 2019-01-25
Does not work with asdf


ionull:
`An unhandled error occurred inside electron-rebuild
No such plugin: nodejs

Failed with exit code: 1

Error: No such plugin: nodejs`

https://github.com/asdf-vm/asdf

PR #273, opened 2019-01-26
fix: Use npm_config_devdir instead of setting HOME

### LINKED ISSUE #117: Use node-gyp's devdir instead of setting env.HOME
I've been attempting to use native modules written in Rust with Electron. My Rust distribution is provided by rustup, which stores Rust distributions and configuration in $HOME. When electron-rebuild calls node-gyp, it overwrites the environment's HOME, which causes rustup to fail to find the rust toolchain.

electron-rebuild sets env.HOME to point node-gyp to the correct Node headers, but this not necessary - if we pass the --devdir argument to node-gyp, then node-gyp will pick up Electron's Node headers, rather than the system's, without breaking any tools that depend on $HOME.

Using versions:
Electron: 1.4.3
electron-rebuild: 1.2.1

fantasticsid:
- changing HOME breaks asdf (https://github.com/asdf-vm/asdf) #271 
- Also should fix #117

### DIFF
--- a/src/rebuild.ts
+++ b/src/rebuild.ts
@@ -358,7 +358,6 @@ class Rebuilder {
     await spawnPromise(nodeGypPath, rebuildArgs, {
       cwd: modulePath,
       env: Object.assign({}, process.env, {
-        HOME: path.resolve(os.homedir(), '.electron-gyp'),
         USERPROFILE: path.resolve(os.homedir(), '.electron-gyp'),
         npm_config_disturl: 'https://atom.io/download/electron',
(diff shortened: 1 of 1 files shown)


fantasticsid:
@ionull I just tested by manually going into my node_modules directory and applied the changes, since it's a short change. Don't know what's the best practice here, but I'm guessing you'd otherwise need to play with yarn/npm so they resolve to a github repo.

MarshallOfSound:
This was published in `1.8.4`

PR #274, opened 2019-02-03
Update CI config

malept:
* Use Ubuntu 16.04 for Linux tests
* Drop Node 4 in AppVeyor
* Remove the Travis workaround, it doesn't seem to be needed anymore
* Increase the timeout on Windows tests only


malept:
Related: #275

PR #278, opened 2019-02-05
chore: use prepare instead of prepublish in package.json scripts

jrobeson:
This should allow installation when installing electron-rebuild from git.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "compile": "tsc",
     "watch": "tsc -w",
-    "prepublish": "npm run compile",
+    "prepare": "npm run compile",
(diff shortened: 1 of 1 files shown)


jrobeson:
This works well for me with npm, but i'm not sure about yarn.

jrobeson:
Yes, this https://github.com/electron/electron-rebuild/commit/dee80faf627d8a3ddc776c73290114ffddba1ef1  . I'm trying to figure out why npm keeps wanting to try use a wrong combination of node-gyp and tar during electron-rebuild.

jrobeson:
Specifically `TypeError: tar.Extract is not a function`

MarshallOfSound:
I'll land this once it's green, but I'll also publish master in a bit 👍

jrobeson:
neat, thanks!

PR #306, opened 2019-06-22
chore: add CFA support

malept:
This is based on the `asar` PR.

Also refactors the "locate electron" code/tests so that they're more robust.


malept:
I gave up with Travis CI's Windows support and went back to AppVeyor.

MarshallOfSound:
`node-gyp` needs to be `^5.0.0` to work on CircleCI (VS2019)

test/electron-locator.ts
```diff
@@ -1,12 +1,12 @@
...
 import * as path from 'path';
 import { expect } from 'chai';
 import { spawnPromise } from 'spawn-rx';
 
-import { locateElectronPrebuilt } from '../src/electron-locater';
+import { locateElectronModule } from '../src/electron-locator';
 
 function packageCommand(command: string, packageName: string) {
-  return spawnPromise('npm', [command, packageName], {
+  return spawnPromise('npm', [command, '--no-save', packageName], {
```
malept: `--no-save` prevents the top-level `package.json` from being overwritten. This is more apparent when developing locally, rather than in CI.

PR #311, opened 2019-07-27
Update node-gyp for VS 2019 support

englercj:
The newest version of node-gyp supports Visual Studio 2019, however the version currently used by this module does not. Without this update windows users cannot use electron-rebuild if they have Visual Studio 2019 installed.


Galkon:
Pls fix :)

Galkon:
Thanks for this PR - need this ASAP though - otherwise we are doing hacky stuff to get builds working.

PR #326, opened 2019-11-02
Updated dependency on node-gyp

### LINKED ISSUE #325: Dependency on node-gyp must be updated
It seems that electron-rebuild depends on node-gyp v5.0.1 which has a bug for macOS Catalina [https://github.com/nodejs/node-gyp/issues/1917](here). The bug has been fixed for the latest version v6.0.1. This bug prevents users from compiling some packages (not all of them), some examples are node-sqlite3, usb amoung others. 

I was able to fix the bug for me simply by updating the dependency on gyp from v5.0.1  to version v6.0.1

daniel3303:
Fixes #325

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "detect-libc": "^1.0.3",
     "fs-extra": "^7.0.1",
     "node-abi": "^2.9.0",
-    "node-gyp": "^5.0.1",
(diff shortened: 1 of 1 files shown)


daniel3303:
Fixes #325

PR #375, opened 2020-07-08
fix: allow "only" modules to match fully qualified scoped module names

issacgerges:
At the moment, modules specified with `-o` or `--only` can only match if you don't supply the prefix. This change allows you to specify a prefixed version of the module but doesn't break the current behavior.

### DIFF
--- a/src/rebuild.ts
+++ b/src/rebuild.ts
@@ -452,7 +452,12 @@ class Rebuilder {
       }
       this.realModulePaths.add(realPath);
 
-      if (this.prodDeps[`${prefix}${modulePath}`] && (!this.onlyModules || this.onlyModules.includes(modulePath))) {
+      if (
+        this.prodDeps[`${prefix}${modulePath}`] &&
+        (!this.onlyModules ||
+          this.onlyModules.includes(modulePath) ||
+          this.onlyModules.includes(`${prefix}${modulePath}`))
+      ) {
         this.rebuilds.push(() => this.rebuildModuleAt(realPath));
       }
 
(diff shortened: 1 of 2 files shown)


test/rebuild.ts
```diff
@@ -144,6 +144,34 @@ describe('rebuilder', () => {
...
+        electronVersion: '5.0.13',
+        arch: process.arch,
+        onlyModules: ['signun'],
+        force: true
+      });
+      let built = 0;
+      rebuilder.lifecycle.on('module-done', () => built++);
+      await rebuilder;
+      expect(built).to.equal(1);
+    });
```
malept, line 173: I think this behavior should be dropped in version 2 - @electron/wg-ecosystem does anyone have differing opinions on this, or know of any legitimate use cases for this behavior?

src/rebuild.ts
```diff
@@ -452,7 +452,12 @@ class Rebuilder {
       }
       this.realModulePaths.add(realPath);
 
-      if (this.prodDeps[`${prefix}${modulePath}`] && (!this.onlyModules || this.onlyModules.includes(modulePath))) {
+      if (
+        this.prodDeps[`${prefix}${modulePath}`] &&
+        (!this.onlyModules ||
+          this.onlyModules.includes(modulePath) ||
```
malept, line 458: This is the line that would also be dropped in version 2.

issacgerges:
I noticed this is still hanging around, just curious if there's anything I can be doing on this one. @MarshallOfSound / @malept.

issacgerges:
Closing as it's been a year

PR #379, opened 2020-08-11
fix: Update to node-gyp 7

adamretter:
node-gyp 7 has some improvements for building on Catalina and building with Python3 etc.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1163,9 +1163,9 @@
       "dev": true
     },
     "chownr": {
-      "version": "1.1.3",
-      "resolved": "https://registry.npmjs.org/chownr/-/chownr-1.1.3.tgz",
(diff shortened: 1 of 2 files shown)


malept:
This will trigger a semver-major release of electron-rebuild, as node-gyp requires a recent version of Node 10. Please hold off on approving this.

adamretter:
@malept Isn't a recent version of Node 10 a good idea? Node10 is already no longer actively developed and will soon be EOL - https://nodejs.org/en/about/releases/

malept:
Yes, however:

1. This needs to be explicit. If this PR was merged as-is, it would be released as a patch, not a major version bump.
2. I have an in-progress branch to update all of the other dependencies that now require Node 10 and I would personally prefer to do all of that together, to avoid multiple version bumps in succession.

malept:
A workaround for now would be to use yarn resolutions in your projects to resolve any node-gyp dependencies to whatever version you need.

malept:
I've upgraded node-gyp to version 7 in the version 2 branch, thanks for letting me know about the version bump.

adamretter:
@malept Is there any sort of timeframe for a version 2 release of electron-rebuild?

malept:
Roadmap: https://github.com/electron/electron-rebuild/issues/356

PR #409, opened 2020-09-09
fix: run node-gyp rebuild correctly

### LINKED ISSUE #404: node-gyp not rebuilding correctly
Native dependency detection appears to be breaking in v2.0.2.

electron-rebuild no longer correctly identifies all the native dependencies needed to be rebuilt.
Reverting back to v2.0.1 fixes the issue.

I will do some more digging into the problem, but the issue is probably related to the changes in 9d166a83bd557084993832c8270f16c688f6ee45

malept:
It should have iterated through all of the node-gyp commands since `node-gyp rebuild` pushes `clean`, `configure`, and `build` commands to the `todo` list. Also:

* ensures that the module path is the current working directory when it's running
* adds a regression test

Fixes #404

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,6 +59,8 @@ jobs:
     executor:
       name: win/vs2019
       shell: bash.exe
+    environment:
+      GYP_MSVS_VERSION: '2019'
     <<: *steps-test
 
   release:
--- a/src/module-rebuilder.ts
+++ b/src/module-rebuilder.ts
@@ -93,8 +93,8 @@ export class ModuleRebuilder {
 
     args.push(...(await this.buildNodeGypArgsFromBinaryField()));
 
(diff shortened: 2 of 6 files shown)


malept:
A further complication: `node-gyp` likes to set the MSVS version (and corresponding environment variable) to 2015, which is very problematic on the CIrcleCI Windows machines. Had to cache the original value when instantiating the Rebuilder class and also restore the environment variable after each call to `rebuild()` in tests.

malept:
@electron/wg-ecosystem this is ready for review. Sorry for the commit noise earlier, I had to figure out the Windows situation.

One person reports that this PR is working for them: https://github.com/electron/electron-rebuild/issues/404#issuecomment-689863252

HashimotoYT (approved):

PR #431, opened 2020-09-30
feat: use electron clang to build native modules

### LINKED ISSUE #432: Use the same compiler that the given Electron version used
In order to maintain maximum compatibility we should build native modules with the same `clang` or `clang-cl` that Electron was built with.

MarshallOfSound:
As in title, this makes it so that when building native modules for Electron you can optionally use the `clang` binary that Electron used to build.  This guarantees support for the same set of compiler features.

Closes #432

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ import rebuild from 'electron-rebuild';
 //     headerURL (optional) - Default: https://www.electronjs.org/headers - The URL to download Electron header files from
 //     types (optional) - Default: ['prod', 'optional'] - The types of modules to rebuild
 //     mode (optional) - The rebuild mode, either 'sequential' or 'parallel' - Default varies per platform (probably shouldn't mess with this one)
+//     useElectronClang (optional) - Whether to use the clang executable that Electron used when building.  This will guaruntee compiler compatibility
 
 // Returns a Promise indicating whether the operation succeeded or not
 ```
--- a/src/clang-fetcher.ts
+++ b/src/clang-fetcher.ts
@@ -1,3 +1,4 @@
+import * as cp from 'child_process';
 import * as debug from 'debug';
 import * as fs from 'fs-extra';
 import got from 'got';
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -135,6 +135,7 @@ import rebuild from 'electron-rebuild';
 //     headerURL (optional) - Default: https://www.electronjs.org/headers - The URL to download Electron header files from
 //     types (optional) - Default: ['prod', 'optional'] - The types of modules to rebuild
 //     mode (optional) - The rebuild mode, either 'sequential' or 'parallel' - Default varies per platform (probably shouldn't mess with this one)
+//     useElectronClang (optional) - Whether to use the clang executable that Electron used when building.  This will guaruntee compiler compatibility
```
malept, line 138: Whether to use the clang executable that Electron used when building its binary.  This will guarantee compiler compatibility

src/clang-fetcher.ts
```diff
@@ -0,0 +1,135 @@
...
+      return fetch(url, responseType, retries - 1);
+    }
+    d('response came back OK');
+    return response.body as any;
+  } catch (err) {
+    d('request failed for some reason', err);
+    await sleep(2000);
+    return fetch(url, responseType, retries - 1);
+  }
+}
```
malept, line 35: Why not at least use [`got`'s `retry` option](https://github.com/sindresorhus/got#retry)? In the exception case, [`async-retry`](https://www.npmjs.com/package/async-retry) or [`p-retry`](https://www.npmjs.com/package/p-retry) would work a bit better because you'd probably want to do something like a logarithmic/exponential backoff.

src/module-rebuilder.ts
```diff
@@ -159,6 +165,24 @@ export class ModuleRebuilder {
     return fs.pathExists(path.resolve(this.modulePath, 'prebuilds', `${process.platform}-${this.rebuilder.arch}`, `electron-${this.rebuilder.ABI}.node`))
   }
 
+  private restoreEnv(env: any) {
```
malept, line 168: Might want to note in a comment that this complicated way to restore the environment is due to Windows.

src/clang-fetcher.ts
```diff
@@ -0,0 +1,135 @@
...
+  d('tar running on clang');
+  await tar.x({
+    file: tarPath,
+    cwd: clangDirPath,
+  });
+  await fs.remove(tarPath);
+  d('cleaning up clang tar file');
+}
+
+if (process.mainModule === module) downloadClangVersion('11.0.0-beta.1')
```
malept, line 135: Was this meant to be left in?

malept (commented):
Post-merge review

src/clang-fetcher.ts
MarshallOfSound: Historically I've had issues with `got`'s retry methodology erroring out instead of retrying.  This is simple enough logic

src/clang-fetcher.ts
MarshallOfSound: Oops

malept:
Hm, I think this PR also undid some of the Dependabot version bump PRs. Example: https://github.com/electron/electron-rebuild/pull/423 and https://github.com/electron/electron-rebuild/pull/441

PR #501, opened 2020-11-18
fix: only call node-abi when there's a module to rebuild

malept:
Electron Forge runs `electron-rebuild` during packaging regardless of whether the Electron app has any native modules (because currently, it doesn't know if it does at that point). This is problematic because that means that `node-abi` is called every time, and if there's a problem with version bumping `node-abi`, they'll incorrectly get the mismatch error.

This pull request should change the rebuilder to only call `node-abi` when `rebuildModuleAt(modulePath)` is called, and cache the result.

### DIFF
--- a/src/rebuild.ts
+++ b/src/rebuild.ts
@@ -47,7 +47,7 @@ const defaultTypes: ModuleType[] = ['prod', 'optional'];
 const ELECTRON_REBUILD_CACHE_ID = 1;
 
 export class Rebuilder {
-  ABIVersion: string | undefined;
+  private ABIVersion: string | undefined;
   nodeGypPath: string;
   prodDeps: Set<string>;
   rebuilds: (() => Promise<void>)[];
--- a/src/rebuild.ts
+++ b/src/rebuild.ts
@@ -47,7 +47,7 @@ const defaultTypes: ModuleType[] = ['prod', 'optional'];
 const ELECTRON_REBUILD_CACHE_ID = 1;
 
 export class Rebuilder {
-  ABI: string;
+  ABIVersion: string | undefined;
   nodeGypPath: string;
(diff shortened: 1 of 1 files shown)


src/rebuild.ts
```diff
@@ -109,13 +109,21 @@ export class Rebuilder {
       throw new Error(`Expected a string version for electron version, got a "${typeof this.electronVersion}"`);
     }
 
-    this.ABI = options.forceABI || nodeAbi.getAbi(this.electronVersion, 'electron');
```
malept, line 112: I have no idea how TypeScript managed to consider this as valid TypeScript, because `forceABI` is a `number` whereas `ABI` is a string.

erickzhao: It seems like `nodeAbi` is of type `any` because we're not using `@types/node-abi`. However, I tried adding `@types/node-abi` in and for some reason, it isn't doing anything. Is there some additional tsc configuration needed?

src/rebuild.ts
```diff
@@ -47,7 +47,7 @@ const defaultTypes: ModuleType[] = ['prod', 'optional'];
 const ELECTRON_REBUILD_CACHE_ID = 1;
 
 export class Rebuilder {
-  ABI: string;
+  ABIVersion: string | undefined;
```
MarshallOfSound, line 50:
```suggestion
  private ABIVersion: string | undefined;
```

src/rebuild.ts
malept: This would be why https://github.com/electron/electron-rebuild/blob/dee80faf627d8a3ddc776c73290114ffddba1ef1/src/ambient.d.ts#L2

Delete that line and it should work?

erickzhao (approved):

PR #525, opened 2020-12-02
chore: use @types/node-abi

erickzhao:
Removes `node-abi` from `ambient.d.ts` and uses the DefinitelyTyped package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "@types/lzma-native": "^4.0.0",
     "@types/mocha": "^8.0.3",
     "@types/node": "^14.6.0",
+    "@types/node-abi": "^2.10.0",
     "@types/tar": "^4.0.3",
     "@types/yargs": "^15.0.5",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -80,7 +81,7 @@
   "eslintConfig": {
     "parser": "@typescript-eslint/parser",
     "parserOptions": {
-      "ecmaVersion": "2018",
+      "ecmaVersion": 2018,
```
erickzhao, line 84: Got this warning in VSCode:

```
Value is not accepted. Valid values: 3, 5, 6, 2015, 7, 2016, 8, 2017, 9, 2018, 10, 2019, 11, 2020, 12, 2021.(1)
```

malept: Not sure why eslint itself didn't complain, but :+1:

malept (approved):
:shipit: assuming tests pass

PR #596, opened 2021-02-01
fix: correctly set the devDir path for node-gyp

k15a:
I am building my electron project with electron-forge and I have some native dependencies. I noticed that Electron Forge is creating a `11.2.0` folder in my project which I traced down to node-gyp not getting the correct devDir.

The root cause seems to be that the `--devdir` option is parsed inside `bin/node-gyp.js` but electron-rebuild is importing `lib/node-gyp.js` directly. This will fix it by setting the `NodeGyp.devDir` property directly.

### DIFF
--- a/src/module-rebuilder.ts
+++ b/src/module-rebuilder.ts
@@ -85,8 +85,7 @@ export class ModuleRebuilder {
       `--target=${this.rebuilder.electronVersion}`,
       `--arch=${this.rebuilder.arch}`,
       `--dist-url=${this.rebuilder.headerURL}`,
-      '--build-from-source',
-      `--devdir="${ELECTRON_GYP_DIR}"`
+      '--build-from-source'
     ];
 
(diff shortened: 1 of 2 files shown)


ckerr:
@k15a please push a commit with an empty [semantic commit message](https://seesparkbox.com/foundry/semantic_commit_messages) to make the CI bots happy :robot:

MarshallOfSound (approved):

k15a:
@ckerr I changed the commit message and the bot seems to be happy now 👍

malept (commented):
This seems fine but it would be nice if we could add a regression test to catch this sort of thing in the future.

k15a:
@malept I did add a small test. Hope this helps.

malept (approved):
Thanks for adding the test! I'll merge this once I commit a readability nit.

Issue #654, opened 2021-03-22
--prebuild-tag-prefix option cannot be empty


Badisi:
I'm trying to rebuild a prebuild package that doesn't have a tag prefix at all.

### prebuild-install (works)

Not using a tag prefix is working fine with `prebuild-install`

```sh
prebuild-install --tag-prefix=''
```

The binary will be downloaded from : `https://github.com/REPO/releases/download/TAG/`

### electron-rebuild (fail)

Not using a tag prefix is not working with `electron-rebuild`

```sh
electron-rebuild --prebuild-tag-prefix=''
```

The url will be : `https://github.com/REPO/releases/download/vTAG/` (where the prefix will still be 'v')
This is due to [cli.ts#L141](https://github.com/electron/electron-rebuild/blob/fcd83a824e9c5bfebc0edebee895676c1347f9d6/src/cli.ts#L141) where the test will fail even if the argument equals an empty string.

PR #655, opened 2021-03-22
fix(#654): --prebuild-tag-prefix option cannot be empty

Badisi:


Badisi:
Any ideas why it failed only on Windows ? because it looks like a CI issue to me..
Maybe someone can just rerun the test ?
Thanks

Badisi:
@MarshallOfSound, I can see that you closed this PR because you renamed `master` branch to `main`, so do I need to resubmit it ? And would you accept it this time ? since you gave no answers to that for about 2 years...
Thanks

PR #703, opened 2021-05-24
build: update clang-fetcher with libcxx headers and object paths

VerteDinde:
The accompanying PR to [29281](https://github.com/electron/electron/pull/29281) in Electron.

Beginning in Electron 13, several nan tests began failing due to an incompatibility between libc++ and libstdc++. This, in turn, could affect any native dependencies built by users using nan. We will now build these native dependencies against libc++ to fix this and prevent future incompatibilities, by copying some of the relevant build arguments from Electron into our nan test runner.

Electron 13 and up now build libc++ and libc++abi as static libraries for Mac and Linux. The relevant zipped headers for these libraries are also checked and uploaded per release - this PR updates `electron-rebuild` with these build changes, and use the zipped headers to apply them.

TODO:
- Test that header path is correct
- Add tests


package.json
```diff
@@ -37,17 +37,19 @@
...
     "debug": "^4.1.1",
     "detect-libc": "^1.0.3",
+    "extract-zip": "^2.0.1",
     "fs-extra": "^9.0.1",
     "got": "^11.7.0",
     "lzma-native": "^7.0.1",
-    "node-abi": "^2.19.2",
-    "node-gyp": "^7.1.0",
+    "node-abi": "^2.30.0",
+    "node-gyp": "^8.0.0",
```
malept, line 50: FYI, if you're going to be upgrading `node-gyp` here, I'd consider this a breaking change since the external dependencies have changed from Python 2 to Python 3.

Also, the version of Python 3 is too new for the existing CI images, so I would suggest cherry-picking this commit: https://github.com/electron/electron-rebuild/pull/669/commits/6d2f8df335d3112a159550dfe6ae9077aafd871c

src/clang-fetcher.ts
```diff
@@ -50,6 +51,45 @@ export async function getClangEnvironmentVars(electronVersion: string, targetArc
...
+      const libcxxObjects = await downloadLibcxxObjects(electronVersion, targetArch)
+      const libcxxHeadersDownloadDir = await downloadLibcxxHeaders(electronVersion, 'libc++')
+      const libcxxabiHeadersDownloadDir = await downloadLibcxxHeaders(electronVersion, 'libc++abi')
+
+      const libcxxHeaders = path.resolve(libcxxHeadersDownloadDir, 'include')
+      const libcxxabiHeaders = path.resolve(libcxxabiHeadersDownloadDir, 'include')
+
+      const cxxflags = [
+        '-nostdinc++',
+        '-D_LIBCPP_HAS_NO_VENDOR_AVAILABILITY_ANNOTATIONS',
```
deepak1556, line 69: FYI: with roll https://github.com/electron/electron/pull/29895 , this macro should be removed in favor of including the `__config_site` file from the root of `libcxx` header file. https://github.com/electron/electron/pull/29895/commits/797723ec838709ddeba0c104e30727ee0b7ac8ca

deepak1556:
Another FYI: libcxx objects built for testing and release are different in terms of the compiler and linker flags used, specifically on linux [ThinLTO](https://source.chromium.org/chromium/chromium/src/+/main:build/config/compiler/compiler.gni;l=65-74) is used. If we don't use the lto flags https://source.chromium.org/chromium/chromium/src/+/main:build/config/compiler/BUILD.gn;l=640-727 to build the native module then miscompile can happen. For ex, compile the following program with the compile and linker flags from this PR, the result will be zero due to miscompilation. Originally caught in https://github.com/microsoft/vscode/pull/129360, I would suggest to find a concrete way to export compiler flags from electron build, probably bundle `libc++.ninja`, `libc++abi.ninja` as part of `libcxx_objects` and this module can extract the values from the file.

```c++
#include <string>
#include <sstream>

int main() {
    std::string message = "Hello";
    uint32_t count = 0;

    std::stringstream final_value;
    final_value << count;
    final_value.write(message.data(), message.size());

    printf("Result : %lu", final_value.str().size());

    return 0;
}
```

package.json
malept: This was resolved in `electron-rebuild@3.0.0`, so the PR needs to be synced with the default branch.

VerteDinde:
Taking another stab at this in light of https://github.com/electron/electron-rebuild/pull/832, may close this PR in favor of a new one.

Issue #718, opened 2021-06-17
Next released version needs to be a major version


malept:
We've merged a few PRs that updated dependencies which require Node.js version 12. Unfortunately, electron-rebuild currently allows Node.js 10 support.

We need to:

* [x] Figure out why CI didn't fail when for the Dependabot PRs that required Node.js 12, despite `--strict-engines`
  * Consider replacing npm (and `package-lock.json`) with Yarn, like a bunch of the other Ecosystem-supported tools.
* [x] Update the required Node.js version in `package.json` (I suggest `>= 12.13.0` to show that we only support LTS)
* [x] Update the CircleCI config to drop Node 10 (and also add Node.js 16 to the Linux CI jobs)
* [x] Update the README to note the minimum Node.js version
* [x] Merge the node-gyp PR
* While we're breaking compatibility:
  * [x] finally drop the sequential options API as per #356
  * [x] Drop support for `electron-prebuilt` (it is incredibly out of date and at this point very insecure)

CC: @ckerr

PR #737, opened 2021-06-29
feat: support `prebuild-install` Node-API modules

### LINKED ISSUE #763: `node-gyp` uses the wrong source of truth to determine `napi_build_version`
I am trying to rebuild the [`keytar@7.7.0`](https://github.com/atom/node-keytar) module for `electron@9.4.4` using `electron-rebuild@1.11.0` and `node@12.22.3`. The issue is that it fails because of what I describe in [another issue](https://github.com/eclipse-theia/theia/pull/9707#issuecomment-880075039):

> * I am using Node 12.22.3 which supports NAPI 8.
> * Our target Electron version is built on Node 12.14.1 which supports NAPI 5.
> * When running `electron-rebuild` which runs `node-gyp` under the hood, it uses my Node to run the cli scripts.
> * `node-gyp` runs `<current-node> -p process.versions.napi` to define the GYP variable `napi_build_version`.
> * `napi_build_version` is set to `8`.
> * C++ build runs, and since we're building for Electron it will use Electron's header files for the C NAPI.
> * `keytar` is made using C++ which has its own header files wrapping the C NAPI.
> * Those C++ headers have a `#if NAPI_VERSION > 5` that controls if it will try to refer to more recent NAPI APIs.
> * Because `NAPI_VERSION` is set to 8 and Electron only support 5, it leads to "missing symbol" errors.

See where `NAPI_VERSION` is set in `node-addon-api` used by `keytar`:  https://github.com/nodejs/node-addon-api/blob/a0c0492f7ceee6a35de9b272ba2231824a61beca/common.gypi#L3

`keytar` also refers to some magic GYP variable `napi_build_version`: https://github.com/atom/node-keytar/blob/af29cf0a17b33038c2c81fd7ed5806fff05c55e3/binding.gyp#L6

---

Is there a way to force `electron-rebuild` and `node-gyp` to define that GYP variable to a specific value, say Electron's supported NAPI version? When going myself into node_modules and changing `"NAPI_VERSION=<(napi_build_version)",` to `"NAPI_VERSION=5",` it works. Alternatively, using a Node version that outputs 5 when doing `node -p process.versions.napi` works too.

timfish:
- Uses [`node-api-version`](https://www.npmjs.com/package/node-api-version) get the max Node-API version for a specific Electron version
  - Even if this modules gets outdated, it will still default to Node-API v8 which will work with all future versions of Electron
- Uses different arguments for `prebuild-install` depending on whether it's a Node-API module
- Throws errors for Electron incompatibility with Node-API modules 

Fixes #763.


src/module-rebuilder.ts
```diff
@@ -105,6 +106,47 @@ export class ModuleRebuilder {
...
+    const binary = (await this.packageJSONFieldWithDefault(
+      'binary',
+      {}
+    )) as Record<string, number[]>;
+
+    if ('napi_versions' in binary) {
+      return binary.napi_versions;
+    }
+
+    return undefined;
```
malept, line 119: How is this different from just `return binary?.napi_versions`?

src/module-rebuilder.ts
```diff
@@ -297,16 +339,26 @@ export class ModuleRebuilder {
...
+    const args = napiVersion
+      ? [
+        '--runtime=napi',
+        `--target=${napiVersion}`,
+      ]
+      : [
+        '--runtime=electron',
+        `--target=${this.rebuilder.electronVersion}`,
+      ];
+
```
malept, line 352: Can we extract this into its own method (and get rid of the ternary), maybe:

```typescript
async getRebuildRuntimeArgs() {
  const napiVersion = await this.getNapiVersion();
  if (napiVersion) {
    return [
      '--runtime=napi',
      `--target=${napiVersion}`,
    ]
  } else {
    return [
      '--runtime=electron',
      `--target=${this.rebuilder.electronVersion}`,
    ]
  }
}
```

src/module-rebuilder.ts
```diff
@@ -297,16 +339,26 @@ export class ModuleRebuilder {
...
       [
         path.resolve(__dirname, `prebuild-shim.${shimExt}`),
         prebuildInstallPath,
         `--arch=${this.rebuilder.arch}`,
         `--platform=${process.platform}`,
-        '--runtime=electron',
-        `--target=${this.rebuilder.electronVersion}`,
-        `--tag-prefix=${this.rebuilder.prebuildTagPrefix}`
+        `--tag-prefix=${this.rebuilder.prebuildTagPrefix}`,
+        ...args,
```
malept, line 361:
```suggestion
        ...await this.getRebuildRuntimeArgs(),
```

May need parentheses, I didn't test this :sweat_smile:

src/module-rebuilder.ts
```diff
@@ -105,6 +106,47 @@ export class ModuleRebuilder {
...
+    const filteredVersions = moduleNapiVersions.filter((v) => v <= electronNapiVersion);
+    
+    if (filteredVersions.length === 0) {
+      throw new Error(`Native module '${this.moduleName}' supports Node-API versions ${moduleNapiVersions} but Electron v${this.rebuilder.electronVersion} only supports Node-API v${electronNapiVersion}`)
+    }
+
+    if (filteredVersions.length === 1) {
+      return filteredVersions[0];
+    }
+
```
malept, line 146: I think this optimization isn't needed.

```suggestion
```

src/module-rebuilder.ts
```diff
@@ -105,6 +106,47 @@ export class ModuleRebuilder {
...
+    const electronNapiVersion = napiVersionFromElectronVersion(this.rebuilder.electronVersion);
+    
+    if (!electronNapiVersion) {
+      throw new Error(`Native module '${this.moduleName}' requires Node-API but Electron v${this.rebuilder.electronVersion} does not support Node-API`);
+    }
+
+    // Filter out Node-API versions that are too high
+    const filteredVersions = moduleNapiVersions.filter((v) => v <= electronNapiVersion);
```
malept, line 137: For clarity, because two double arrows makes it kind of strange to read:
```suggestion
    const filteredVersions = moduleNapiVersions.filter((v) => (v <= electronNapiVersion));
```

src/module-rebuilder.ts
```diff
@@ -105,6 +106,47 @@ export class ModuleRebuilder {
...
+
+    return undefined;
+  }
+
+  async getNapiVersion(): Promise<number | undefined> {
+    const moduleNapiVersions = await this.getSupportedNapiVersions();
+
+    if (!moduleNapiVersions) {
+      // This is not a Node-API module
+      return undefined;
```
malept, line 127: I wonder if `return` is more idiomatic than `return undefined`?

malept (changes requested):
I think this is a good approach, I have some questions/suggestions about the implementation.

Also, this change needs tests.

test/rebuild.ts
```diff
@@ -77,6 +77,10 @@ describe('rebuilder', () => {
         await expectNativeModuleToBeRebuilt(testModulePath, 'farmhash');
       });
 
+      it('should have rebuilt top level prod dependencies that are using prebuild + napi', async () => {
+        await expectNativeModuleToBeRebuilt(testModulePath, 'node-hid');
+      });
+
```
malept, line 83: I took a quick look, `farmhash` uses N-API so that case is already covered. I think what I'm more interested in, in terms of integration tests, is when farmhash is "rebuilt", it doesn't hit the `node-gyp` codepath.

malept (changes requested):
Sorry, I should have been more specific about testing.

In terms of unit tests, I'd like to see some fixture-based tests that exercise the edge cases such as the new thrown errors. For example, create fake native modules (no need for actual C++ code, just the proper `package.json` prebuild/N-API config) that test unsupported N-API versions and Electron versions that don't support N-API.

test/rebuild.ts
timfish: Right, got it!

timfish:
Not 100% sure why `should rebuild only specified modules` keeps failing. When that test was using `farmhash` it was failing because the hash didn't change between rebuilds. I'm guessing this is because a prebuilt binary is now getting correctly downloaded. I've changed that test to `bcrypt` and not its failing on some platforms due to not finding the binary.

malept (commented):
For the record, I'm going to hold on this PR until #718 is resolved, due to the weird state of the default branch. But otherwise this PR seems good to go.

malept:
Closing in favor of #737 because I screwed up the merge of v3.0.0.

PR #784, opened 2021-07-30
chore: update deps, require Node 12

MarshallOfSound:
As in title, updates a bunch of dependencies.  To do so easily I moved us to yarn via `yarn import`

To make CI work this also formalizes the bump to require node 12 as indicated by #718


test/electron-locator.ts
```diff
@@ -9,7 +9,7 @@ import { locateElectronModule } from '../src/electron-locator';
 const testElectronRange = require(path.resolve(__dirname, '..', 'package.json')).devDependencies.electron;
 
 function packageCommand(command: string, packageName: string): Promise<string> {
-  return spawn('npm', [command, '--no-save', packageName], {
+  return spawn('npm', [command, '--no-save', '--no-package-lock', packageName], {
```
malept, line 12: I reverted the yarn change here because it seems that yarn doesn't have a concept of "don't actually update the package.json for this".

.circleci/config.yml
```diff
@@ -15,34 +15,30 @@ steps-test: &steps-test
...
+              nvm install 12
+              nvm alias default 12
               echo 'export NVM_DIR=${HOME}/.nvm' >> $BASH_ENV
               echo "[ -s '${NVM_DIR}/nvm.sh' ] && . '${NVM_DIR}/nvm.sh'" >> $BASH_ENV
               ;;
             Windows*|CYGWIN*|MINGW*|MSYS*)
-              nvm install 10.22.0
-              nvm use 10.22.0
+              nvm install 12.22.1
+              nvm use 12.22.1
```
malept, line 25: The Windows CI worker is giving some extremely strange errors, going to try a different version here
```suggestion
              nvm install 12.22.4
              nvm use 12.22.4
```

test/electron-locator.ts
malept: Update: I completely removed the package manager code in favor of fixtures.

PR #799, opened 2021-08-09
feat!: 3.0.0

### LINKED ISSUE #718: Next released version needs to be a major version

malept:
BREAKING CHANGE: deprecated APIs are removed

BREAKING CHANGE: remove support for `electron-prebuilt`, as it's super old and unsupported

BREAKING CHANGE: Requires Node 12 LTS

Also adds `RebuildOptions` to the top-level API.

Fixes #718.


ckerr (approved):
Looks pretty straightforward to me

PR #800, opened 2021-08-09
feat: support `prebuild-install` for Node-API modules

### LINKED ISSUE #520: electron-rebuild breaks prebuild-install, sends both `--runame=napi` and `--runtime=electron` args
The latest release of `node-hid` (2.0.0-1) uses N-API instead of NAN.  The proper `prebuild-install` invocation for N-API modules is 
```
prebuild-install --runtime napi 
```
but when doing `electron-rebuild` in a project, like [electron-hid-test](https://github.com/todbot/electron-hid-test), I'm seeing instead that it's being invoked like:
```
./node_modules/node-hid/node_modules/.bin/prebuild-install --runtime=napi --target 11.0.3 --runtime=electron --verbose
```
and `prebuild-install` errors out with `Error: Could not detect abi for version 11.0.3 and runtime napi,electron`.

### LINKED ISSUE #763: `node-gyp` uses the wrong source of truth to determine `napi_build_version`

malept:
Continuation of #737 (because I screwed up the merge)

Fixes #520.
Fixes #763.


malept:
CC: @timfish

PR #801, opened 2021-08-09
chore: update codecov image URL due to default branch change

malept:
Don't merge until the default branch changes.  AFAIK this is the only change needed.


dsanders11:
Branch name was changed (in addition to a repo rename) so gonna close this as it's no longer needed.

PR #850, opened 2021-09-02
rebuild: include prefix when filtering with --only

paul-marechal:
The logic currently ignores a module prefix when filtering packages
passed via `--only=`. This is quite confusing: to only rebuild
something like `@theia/node-pty` we must pass `--only=node-pty`.

This commit adds a lookup using the prefixed name.

### DIFF
--- a/src/rebuild.ts
+++ b/src/rebuild.ts
@@ -324,7 +324,7 @@ export class Rebuilder {
       }
       this.realModulePaths.add(realPath);
 
-      if (this.prodDeps[`${prefix}${modulePath}`] && (!this.onlyModules || this.onlyModules.includes(modulePath))) {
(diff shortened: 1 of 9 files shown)


paul-marechal:
I'm not sure if I should just replace `.includes(modulePath)` by ``.includes(`${prefix}{modulePath}`)`` as it might be breaking?

malept (changes requested):
I think this needs tests for regression purposes.

paul-marechal:
I added tests, but I'll need to close this PR temporarily, I will re-open when ready.

paul-marechal:
Ready to go again.

erickzhao:
drafting due to conflicts. i think i can work on resolving them and getting this shipped.

PR #888, opened 2021-10-08
fix: enforce extraModules to be a set of strings

paul-marechal:
For some reason TypeScript is AOK with stuffing `Set` objects with
random keys. This is problematic as this is not how `Set` are supposed
to be used (like a plain JavaScript object map).

Fix the logic to respect the `-w/--which-module` CLI argument.


ckerr (approved):
Code LGTM. Approved, but I'll hold off merging since @malept was pinged in case he wants to review.

@paul-marechal please ping me if nothing happens in a week :smile_cat:

paul-marechal:
@ckerr @malept updated the tests. I modified one of the tests to delete all dependencies from the package.json file and make use of `extraModules`.

vince-fugnitto:
@malept @ckerr any updates?

malept (commented):
It seems to me like there should be a better way to test this. I need to think about it.

paul-marechal:
> Can the added unit tests not cause modules to be rebuilt? Trying not to increase the test time when possible.

I did it this way because of the above comment, I thought the new test I had added was causing issues...

With the current way, it won't build (many) more packages than what it used before, keeping the test run time low. Moreover the specific test modified is specifically trying to rebuild packages that aren't part of dependencies, so this hackish way of modifying the package.json just for this case seemed reasonable to me.

If the test is that problematic I can always remove it :)

The previous logic was broken without a doubt, but the fix is trivial.

paul-marechal (commented):
Your last commit looks good to me! Thanks for helping with this, I didn't feel familiar enough with the internals to do what you did.

paul-marechal:
@malept should I do anything?

vince-fugnitto:
@malept any update?

vince-fugnitto:
ping.

tsmaeder:
Hello, is this thing on? The PR has been open for 2.5 years and approved for half a year. Can we get a decision, please?

erikian (approved):
Looks like all feedback was addressed, so this should be good to land

PR #907, opened 2021-11-08
fix: require node-gyp@^8.4.0 & handle old Electron versions

zcbenz:
Since node-gyp >= 8.4.0, building modules for old versions of Electron requires passing `--force-process-config` due to them lacking a valid config.gypi in their headers.

See also https://github.com/nodejs/node-gyp/pull/2497.

### DIFF
--- a/src/module-type/node-gyp.ts
+++ b/src/module-type/node-gyp.ts
@@ -3,6 +3,7 @@ import detectLibc from 'detect-libc';
 import NodeGypRunner from 'node-gyp';
 import path from 'path';
 import { promisify } from 'util';
+import semver from 'semver';
 
 import { ELECTRON_GYP_DIR } from '../constants';
 import { getClangEnvironmentVars } from '../clang-fetcher';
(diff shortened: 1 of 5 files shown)


malept (commented):
Can we start adding tests for `src/module-type/node-gyp.ts`? You'd be adding `test/module-type-node-gyp.ts` and testing when `--force-process-config` is added to the args.

malept (changes requested):
Going to actually block on the testing request.

zcbenz:
I have updated the code to add tests.

src/module-type/node-gyp.ts
```diff
@@ -36,6 +36,17 @@ export class NodeGyp extends NativeModule {
...
     }
 
+    // Headers of old Electron versions do not have a valid config.gypi file
+    // and --force-process-config must be passed to node-gyp >= 8.4.0 to
+    // correctly build modules for them.
+    // See also https://github.com/nodejs/node-gyp/pull/2497
+    const [major, minor] = this.rebuilder.electronVersion.split('.').slice(0, 2).map(n => parseInt(n, 10));
+    if ((major <= 13) ||
+        (major == 14 && minor <= 1) ||
+        (major == 15 && minor <= 2)) {
```
malept, line 46: @zcbenz what do you think of using `semver` here instead, so that this is more readable? It's already a transitive dependency due to `node-abi`, among other dependencies. Proposed code (assuming I have the logic correct):
```suggestion
    if (!semver.satisfies(this.rebuilder.electronVersion, '^14.2.0 || ^15.3.0 || >= 16.0.0-alpha.1')) {

```

I made the change locally and tests still pass.

malept (commented):
I've refactored the tests so that they should test all of the corner cases (and sped them up). But I have an implementation question which could potentially make this a bit easier to maintain.

src/module-type/node-gyp.ts
zcbenz: Using `semver` makes the code looks much better!

malept (approved):

PR #1022, opened 2022-07-06
fix: add std=c++17 flag for e20+

VerteDinde:
Addresses e/e issue [34209](https://github.com/electron/electron/issues/34209) while 20.0.0-beta.1 is in beta.

This PR adds a compile flag for `-std=c++17` when using Electron 20 or higher.

### DIFF
--- a/src/clang-fetcher.ts
+++ b/src/clang-fetcher.ts
@@ -2,6 +2,7 @@ import * as cp from 'child_process';
 import debug from 'debug';
 import * as fs from 'fs-extra';
 import * as path from 'path';
+import semver from 'semver';
 import * as tar from 'tar';
 import * as zlib from 'zlib';
 import { ELECTRON_GYP_DIR } from './constants';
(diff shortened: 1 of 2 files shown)


malept:
Is this a `chore:` or a `fix:`?

src/clang-fetcher.ts
```diff
@@ -40,6 +40,9 @@ export async function getClangEnvironmentVars(electronVersion: string, targetArc
   if (process.platform === 'darwin') {
     clangArgs.push('-isysroot', getSDKRoot());
   }
+  if (parseInt(electronVersion, 10) >= 20) {
```
malept, line 43: suggestion: `semver` is a direct dependency, might as well use it?

malept (commented):
Might be worth adding a test if we have unit tests for `clang-fetcher`?

VerteDinde:
@malept Great point - updated commit message

VerteDinde:
@malept Looks like we don't have any unit tests for clang-fetcher - I'll see if I can add some

ckerr (approved):
This is good for now and solves a real problem.

For `electron/electron` I'm working on some solution that would provide the build flags that come from upstream so that we don't have to hardcode a particular version. *However* in practice this is only going to change every few years (any predictions on when Chromium will jump to C++20?)

Also, upstream hardcodes this in several places too, e.g. https://chromium-review.googlesource.com/c/chromium/src/+/3306812 for one example.

Issue #1024, opened 2022-07-19
v3.2.8 breaks packaging electron app with ffi-napi dependency


t57ser:
Following error is thrown:

An unhandled rejection has occurred inside Forge:
Error: /m/i1/temp/buildTmp/electron-packager/linux-x64/test-x64/resources/app/node_modules/ffi-napi/build/node_gyp_bins/python3: file links out of the package

Using electron-forge 6.0.0-beta.64 with electron 20 beta and ffi-napi 4.0.3
Only happens on mac and linux

Once i revert back to 3.2.7 packaging the application is working again.

VerteDinde:
@t57ser Can you include the module that's breaking, or a repro of the break in Electron Fiddle? We're able to build a module with ff-napi without seeing this error, just want to try and get a repro for a deeper look 👀

t57ser:
@VerteDinde Here you go: https://gist.github.com/t57ser/3be2bc0d590690ece381f69e8c316edf
remember this only happens on mac, not on windows

```
An unhandled rejection has occurred inside Forge:
Error: /var/folders/sp/141453k53mg70282wr9vkn_h0000gp/T/electron-packager/darwin-x64/test-darwin-x64/Electron.app/Contents/Resources/app/node_modules/ffi-napi/build/node_gyp_bins/python3: file links out of the package

```

VerteDinde:
Sorry @t57ser, could you include the Mac/Linux arch you're using, and the version of Electron 20-beta? I've been trying to repro on Mac arm64 using your gist, and I haven't been able to hit this error. We have another maintainer on Linux who was also trying to repro on Ubuntu 22.04 / Xorg (using Electron 20.0.0-b13 and  electron-rebuild 3.2.8) who is also not able to repro. 

Maybe this is specific to a Python version or arch? Just trying to get more details so we can try to fix this before Electron 20 stable for you 🤔

t57ser:
macOS Monterey 12.4. x86_64
electron 20.0.0-beta.11
I am quite surprised you can not reproduce this since i have tried this on multiple machines
I just tried a brand new mac, installed node etc... and same error

t57ser:
Starting electron works, but packaging does not work, maybe that is the misunderstanding.
So in my example that i send, you need to `npm run package`

fcastilloec:
I'm having the same issue, trying to compile for electron 16.
I can see the issue in Linux and macOS, I haven't tested Windows yet.
[...]

PR #1025, opened 2022-08-01
fix: remove build/node_gyp_bins if it exists

VerteDinde:
Addresses #1024

In node-gyp v9.0.0, node-gyp adds a python symlink for older versions of python. This symlink is added to a `node_gyp_bins/python3` directory within the build directory, which electron-rebuild then tries to rebuild. This PR removes the `node_gyp_bins` directory once the node gyp module has finished rebuilding.

### DIFF
--- a/src/module-rebuilder.ts
+++ b/src/module-rebuilder.ts
@@ -93,6 +93,7 @@ export class ModuleRebuilder {
     await this.nodeGyp.rebuildModule();
     d('built via node-gyp:', this.nodeGyp.moduleName);
     await this.writeMetadata();
+    await this.removeNodeGypBins();
     await this.replaceExistingNativeModule();
     await this.cacheModuleState(cacheKey);
     return true;
@@ -117,6 +118,16 @@ export class ModuleRebuilder {
         await fs.copyFile(nodePath, path.join(abiPath, `${this.nodeGyp.moduleName}.node`));
       }
     }
+  }
+
+  /**
+   * Removes symlinks created by node-gyp for external dependencies, such as Python.
+   * We do not want to include these in rebuilt packaged apps.
+   */
+  async removeNodeGypBins(): Promise<void> {
+    const nodeGypBinsLocation = path.resolve(this.modulePath, 'build', 'node_gyp_bins');
+    d('removing node_gyp_bins in ', nodeGypBinsLocation);
+    await fs.remove(nodeGypBinsLocation)
   }
 
   async writeMetadata(): Promise<void> {


src/module-rebuilder.ts
```diff
@@ -119,6 +120,23 @@ export class ModuleRebuilder {
...
+   * We do not want to include these in rebuilt, packaged apps.
+   */
+  async removeNodeGypBins(): Promise<void> {
+    const buildLocation = path.resolve(this.modulePath, 'build');
+
+    d('checking existance of node_gyp_bins', buildLocation);
+    const nodeGypBinsLocation = path.resolve(buildLocation, 'node_gyp_bins');
+    const exists = await fs.pathExists(nodeGypBinsLocation);
+  
+    if (exists) {
```
malept, line 134: question: what are the downsides to putting the await in the conditional?

src/module-rebuilder.ts
```diff
@@ -119,6 +120,23 @@ export class ModuleRebuilder {
...
   }
 
+  /**
+   * Removes symlinks created by node-gyp for external dependencies, such as Python.
+   * We do not want to include these in rebuilt, packaged apps.
+   */
+  async removeNodeGypBins(): Promise<void> {
+    const buildLocation = path.resolve(this.modulePath, 'build');
+
+    d('checking existance of node_gyp_bins', buildLocation);
```
malept, line 130: Nit:

```suggestion
    d('checking for the existence of node_gyp_bins', buildLocation);
```

malept (commented):
This seems fine from a high level.

Can there be a test for this new code?

src/module-rebuilder.ts
VerteDinde: D'oh, dumb typo, thanks! Will fix 👍

src/module-rebuilder.ts
VerteDinde: Fair question - reading the fs docs, it looks like `fs.remove` will either remove the directory if it exists, or just silently continue, so I think the conditional here is actually unnecessary. I'll remove it 👍

VerteDinde:
@malept Definitely - I pushed this up as a draft first to make sure it passed CI, looking at how to best test this now

malept:
Whoops, I missed the draft part. Sorry about that.

MarshallOfSound (changes requested):
Hot take, this _can_ go here, but I think it probably belongs in some kind of default `ignore` list in packager 🤔 

 Can we also potentially follow-up upstream with a change that moves this directory outside of `node_modules`.  I don't see a technical need for the `node_gyp_bins` folder to be generated for every module or in the actual module folder itself (vs just being in a tmpdir)

VerteDinde:
@MarshallOfSound Yeah, I think that's a fair take! 🙂 I'll close this one and move the work over to packager. Will also file an issue upstream with node-gyp itself

PR #1027, opened 2022-08-18
feat: add support for pnpm workspace

seokju-na:
Hello ✋ 

When using the pnpm workspace, there was a problem that some native modules were missing from the build target because the project root path is incorrect.

This PR adds support for pnpm workspace by find where `pnpm-lock.yaml` is located.


BlackHole1 (approved):
LGTM. Thanks to @seokju-na's PR and @hipstersmoothie's reminder.

dsanders11 (approved):
Made a minor change to correct the filename, which the PR description lists as `pnpm-lock.yaml` but was originally in the code as `pnpm-lock.json` - confirmed the former is correct and updated the code.

PR #1060, opened 2023-01-06
build: use cfa orb

erickzhao:


dsanders11:
@erickzhao, I ran into the `yarn` not found error on Windows on another CI config, it's related to bumping to `circleci/windows@5.0.0`. If this can still use an older version of that orb, it'll unblock that issue.

dsanders11:
Closing as this was implemented in #1086.

PR #1072, opened 2023-03-07
fix: cross-platform build

cocktailpeanut:
The rebuilder has been using the hard-coded `process.platform` value as the platform attribute all this time.

This means, no matter what platform option you pass in, it would just use the current OS platform.

Basically, cross platform prebuild has not been working at all.

Combined with the electron-builder fix here https://github.com/cocktailpeanut/electron-builder/commit/73f751b42f9b7a9b636d9ae39fc340d6b2890c3b - now you can do cross platform electron builds that include prebuilt binaries.

### DIFF
--- a/src/rebuild.ts
+++ b/src/rebuild.ts
@@ -14,6 +14,7 @@ export interface RebuildOptions {
   buildPath: string;
   electronVersion: string;
   arch?: string;
+  platform?: string;
   extraModules?: string[];
   onlyModules?: string[] | null;
   force?: boolean;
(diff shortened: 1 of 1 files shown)


MarshallOfSound (changes requested):
I think this is valid for prebuild modules but is invalid for modules powered via node-gyp.  There should probably be checks in place to ensure that non-system-platform builds aren't even attempted for node-gyp builds

s00d:
The fix is needed for controlling the OS during the assembly process. For example, there is currently an issue when using electron-builder version 24, which started using electron/rebuild. The problem can be found here: [link to the specific line of code on GitHub](https://github.com/electron-userland/electron-builder/blob/46524169cefbfa18e342d7fa19e79e710aae848e/packages/app-builder-lib/src/util/yarn.ts#L162).

The OS will always be the current one, and because of this, it is not possible to download the necessary bindings.

Please merge this fix.

tmm1:
Looks like this was resolved in https://github.com/electron/rebuild/pull/1153

PR #1085, opened 2023-06-13
ci: Test current versions of Node.js

cclauss:
Builds on #1084 to have CircleCI test on the versions of Node.js at https://github.com/nodejs/release#readme

Linux 20 fails on:
```
[4/4] Building fresh packages...
$ npm run compile
/usr/bin/env: ‘node’: Text file busy
error Command failed with exit code 126.
```

PR #1086, opened 2023-06-14
ci: update test workflow and expand test matrix

dsanders11:
Supersedes #1060
Supersedes #1085 (this PR tests a larger matrix of versions, including on macOS and Windows)

This was a real headache of conflicting bugs but managed to get a config that goes green for the full matrix:
* Tests against Node.js 20.2.0 because 20.3.0 has a bug (see comment in #1085) - also works around a panic in `nvm-windows` when installing 20.3.0
* Uses the `win/server-2022` executor because `win/vs2019` is no longer updated and has an old `nvm-windows` version which has a bug with npm versions
* [Enables long paths on Windows](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later) because with the `win/server-2022` executor the path of files being compiled ended up being too long 🤷
* The Windows builds are unfortunately flaky due to hitting the race condition in https://github.com/nodejs/node-gyp/pull/2846 and I don't have a good fix at the moment - ideally we should work around that bug in this project since it will be a while before the `node-gyp` fix makes its way to most users

cc @cclauss, sorry to supersede your PR (but thanks for putting in the work!), but I've been meaning to update the CI config here for a while, these changes pull it closer to our other @~electron/wg-ecosystem repos and pick up other needed changes from #1060.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,99 +1,132 @@
-step-restore-cache: &step-restore-cache
-  restore_cache:
-    keys:
-      - v1-dependencies-{{ arch }}-{{ checksum "yarn.lock" }}
-
-steps-test: &steps-test
-  steps:
-    - checkout
-    - *step-restore-cache
-    - run:
-        name: Install Node
-        command: |
-          case "$(uname -s)" in
-            Darwin)
-              curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.34.0/install.sh | bash
-              export NVM_DIR="$HOME/.nvm"
-              [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
-              nvm install 12
-              nvm alias default 12
-              echo 'export NVM_DIR=${HOME}/.nvm' >> $BASH_ENV
-              echo "[ -s '${NVM_DIR}/nvm.sh' ] && . '${NVM_DIR}/nvm.sh'" >> $BASH_ENV
-              ;;
-            Windows*|CYGWIN*|MINGW*|MSYS*)
-              nvm install 12.22.4
-              nvm use 12.22.4
-              ;;
-          esac
-    - run: yarn install --frozen-lockfile --ignore-engines
-    - save_cache:
-        paths:
-          - node_modules
-        key: v1-dependencies-{{ arch }}-{{ checksum "yarn.lock" }}
-    - run: yarn run lint
-    - run:
-        name: Tests with code coverage
-        command: yarn run coverage
-        environment:
-          DEBUG: electron-rebuild
-    - run: yarn run codecov
-
-
 version: 2.1
+
 orbs:
-  win: circleci/windows@1.0.0
+  cfa: continuousauth/npm@1.0.2
+  node: circleci/node@5.1.0
+  win: circleci/windows@5.0.0
+
+commands:
+  install:
+    parameters:
+      node-version:
+        description: Node.js version to install
+        type: string
+    steps:
+      - run: git config --global core.autocrlf input
+      - node/install:
+          node-version: << parameters.node-version >>
+      - run: nvm use << parameters.node-version >>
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1,99 +1,132 @@
...
-      - release:
+      - test-linux:
+          matrix:
+            parameters:
+              node-version:
+                - 20.2.0
+                - 18.16.0
+                - 16.20.0
+                - 14.21.3
+                - 12.22.12
```
cclauss, line 94: Why continue to support and encourage the use of end-of-life versions of Node.js?  If users want support then they can try to reproduce their problems on a supported version.  If folks are paying for long-term support for an OS that has an EOL version of Node.js then their OS provider can support them. If folks are using an OS that is beyond the standard support and are not paying for LTS, then their problems should be demonstrated on supported versions.  Building on secure platforms and being able to use new language optimizations and features helps everyone's safety and velocity so let's encourage security and performance upgrades.

dsanders11: This package has an engine version that supports Node.js v12 and above, so dropping support would be a breaking change (and require a major version bump on this package). Any major version bump causes fragmentation with the user base, as can be seen [in the npm downloads stats for older versions of this package](https://www.npmjs.com/package/electron-rebuild?activeTab=versions) (it used to be called just `electron-rebuild`) where there are still ~6k downloads/week on the final 2.y.z version and ~7.5k downloads/week on the final 1.y.z version. We want to avoid furthering that fragmentation, so we don't want to bump majors more often than we need to (which every Node.js EOL would require). These packages also don't have a ton of maintainer time to put into them, so it takes some time for things to be merged - if we bump the major version we might make the safety issue worse by putting new fixes behind a new major version where they won't be picked up by some subset of the user base. For ease of maintenance these packages aren't setup to do additional releases to old majors, so once we bump majors there are no more fixes being backported to old majors (and we would need the CI to be testing the old Node.js versions those majors support to be able to do so).

So the general policy of @~electron/wg-ecosystem is to continue to test and not explicitly drop older Node.js versions unless necessary. We're not encouraging the usage of EOL versions of Node.js, more avoiding breaking them when feasible. In most cases we won't go out of our way to fix bugs which only affect those EOL versions, and if a future change requires dropping support for older EOL versions of Node.js we will explicitly drop support then, if it's deemed necessary.

Hope this makes sense. 🙂

cclauss: Users of legacy software create a lot of noise in support channels...
* 123 issues: https://github.com/search?q=repo%3Anodejs%2Fnode-gyp+electron-rebuild&type=issues

dsanders11: Will that noise really go away if this package stops testing those EOL Node.js versions, though? Even if we drop support and bump the major, some users will stick on the current major (as seen in the npm download stats listed above).

If you tweak the GitHub search to include `created:>2022-01-01`, it only returns 16 issues in the past 18 months, and without going beyond flipping through the descriptions shown on the search page, at least 3 of them were reporting supported Node.js versions.

It's also worth noting that `node-gyp` itself is still showing Node.js v12 and v14 as supported engines according to `package.json`: https://github.com/nodejs/node-gyp/blob/33391db3a0008eff8408890da6ab232f2f90fcab/package.json#L37-L39

cclauss: > Will that noise really go away if this package stops testing those EOL Node.js versions?

Rephrase: Will that noise **decrease** if this package stops **encouraging / supporting** EOL Node.js versions?

Yes, it will.  Developers are improving software and ignoring their efforts is suboptimal on many different levels.  Solutions that used to work in older versions of npm, node, and gyp no longer work today.  Supporting both legacy and current is complex, frustrating, and error-prone.  Doing so for the sake of statistics is a shame.

cclauss (commented):
LGTM: Nice work here!

erickzhao (approved):

PR #1095, opened 2023-08-02
feat: add node-pre-gyp support

### LINKED ISSUE #367: node-pre-gyp support in the same way than prebuild is supported
@MarshallOfSound @malept pinging you folks as you are the most recent contributors, sorry to bother with this, but after looking for a similar solution I found this PR which was not accepted: https://github.com/electron/electron-rebuild/pull/170.

`node-pre-gyp` does allow you to have prebuilt binaries for electron, just like `prebuilt`. For a working example, see the assets on this release: https://github.com/JCMais/node-libcurl/releases/tag/v2.1.2-1

The trick is that to make it use the correct prebuilt binary some arguments / environment variables must be passed with the `npm install` / `yarn install` command, like:
```bash
# npm
npm install node-libcurl --runtime=electron --target=v8.2.5 --disturl=https://www.electronjs.org/headers --save

# yarn
npm_config_runtime=electron npm_config_target=v8.2.5 npm_config_disturl=https://www.electronjs.org/headers yarn add node-libcurl
```

If those flags are not passed the prebuilt binary for the current Node.js version is going to be downloaded, and then later when `electron-rebuild` is called the addon is built from source, instead of trying to use the prebuilt binary for Electron.

Is the project willing to accept a PR that adds support to it? There is no need to bring the whole `node-pre-gyp` dependency itself as the linked PR did, as the process should be similar to the one used to support `prebuild`:
https://github.com/electron/electron-rebuild/blob/c42d6c7ed3ea07b6203a686ccdbfbae6bd5f8b5f/src/rebuild.ts#L295-L336

There are multiple related issues on this repo about `node-pre-gyp` support, the most recent one is this one:
https://github.com/electron/electron-rebuild/issues/309

I should be able to dedicate some time to this PR, but I need to know if there is any possibility of it being accepted.

samuelmaddock:
[@mapbox/node-pre-gyp](https://www.npmjs.com/package/@mapbox/node-pre-gyp) is a popular choice for serving prebuilt native node module binaries. These changes add support for downloading prebuilt binaries using it as well as rebuilding in the case that no prebuilt binaries are found.

A `skipPrebuilds` option was added to give tests the ability to force rebuilds without using any prebuilt binary modules.

[node-sqlite3](https://github.com/TryGhost/node-sqlite3) is used to test for node-pre-gyp support.

resolves #367

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -128,6 +128,7 @@ process.on('unhandledRejection', handler);
     useElectronClang: !!argv.useElectronClang,
     disablePreGypCopy: !!argv.disablePreGypCopy,
     projectRootPath,
+    skipPrebuilds: !!argv.skipPrebuilds,
   });
 
   const lifecycle = rebuilder.lifecycle;
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -128,7 +128,7 @@ process.on('unhandledRejection', handler);
     useElectronClang: !!argv.useElectronClang,
     disablePreGypCopy: !!argv.disablePreGypCopy,
     projectRootPath,
-    skipPrebuilds: !!argv.skipPrebuilds,
+    buildFromSource: !!argv.buildFromSource,
   });
 
   const lifecycle = rebuilder.lifecycle;
--- a/test/helpers/module-setup.ts
+++ b/test/helpers/module-setup.ts
@@ -35,7 +35,7 @@ export async function resetTestModule(testModulePath: string, installModules = t
 }
 
 export async function cleanupTestModule(testModulePath: string): Promise<void> {
-  await fs.remove(testModulePath);
+  await fs.rmdir(testModulePath, { recursive: true, maxRetries: 10 });
   resetMSVSVersion();
 }
 
--- a/test/module-type-prebuild-install.ts
+++ b/test/module-type-prebuild-install.ts
@@ -35,7 +35,10 @@ describe('prebuild-install', () => {
       ])
     });
 
-    it('should not fail running prebuild-install', async () => {
+    it('should not fail running prebuild-install', async function () {
+      if (process.platform === 'darwin' && process.arch === 'arm64') {
+        this.skip(); // farmhash module has no prebuilt binaries for ARM64
+      }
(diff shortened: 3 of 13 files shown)


src/cli.ts
```diff
@@ -128,6 +128,7 @@ process.on('unhandledRejection', handler);
     useElectronClang: !!argv.useElectronClang,
     disablePreGypCopy: !!argv.disablePreGypCopy,
     projectRootPath,
+    skipPrebuilds: !!argv.skipPrebuilds,
```
MarshallOfSound, line 131: Does this new option also impact `prebuild` or just `node-pre-gyp`?

MarshallOfSound: It also appears to have a `buildFromSource` behavior rather than "skipping" the module entirely

georgexu99: renamed `prebuild` -> `buildFromSource`

VerteDinde (approved):
The code changes look good to me

test/helpers/module-setup.ts
```diff
@@ -35,7 +35,7 @@ export async function resetTestModule(testModulePath: string, installModules = t
 }
 
 export async function cleanupTestModule(testModulePath: string): Promise<void> {
-  await fs.remove(testModulePath);
+  await fs.rmdir(testModulePath, { recursive: true, maxRetries: 10 });
```
MarshallOfSound, line 38: ?

test/module-type-prebuild-install.ts
```diff
@@ -35,7 +35,10 @@ describe('prebuild-install', () => {
       ])
     });
 
-    it('should not fail running prebuild-install', async () => {
+    it('should not fail running prebuild-install', async function () {
+      if (process.platform === 'darwin' && process.arch === 'arm64') {
+        this.skip(); // farmhash module has no prebuilt binaries for ARM64
```
MarshallOfSound, line 40:
```suggestion
        this.skip(); // farmhash module has no prebuilt binaries for ARM64
        return;
```

georgexu99:
o shoot, I merged before sam's comments loaded :PPP I'll make a new pr to address these

Issue #303, opened 2019-06-05
Expose node-gyp's jobs For Multi Core Building


dsanders11:
Currently `electron-rebuild` uses a single `node-gyp` job (the default) so it can only use a single core. With multi-core systems (should be almost all these days) this leaves a lot of performance on the table and results in slow builds.

The fix is easy enough, `jobs N` simply needs to be passed to `node-gyp`. I'll gladly make a PR, but wanted to discuss how it should look before putting any time into it.

Tested the change manually and on a rebuild on a Raspberry Pi, build time went from 1409 seconds to 693 seconds because it was able to use more cores. In theory it should have only taken ~350 seconds but I think it was I/O limited due to a slow disk.

The most obvious approach for exposing this option would be to add `-j, --jobs   Number of parallel jobs to use` as an option for `electron-rebuild`, and pass this value to `node-gyp`. This would retain current behavior of only using a single job, but allow configuring more.

The downside to this approach is there's no automatic performance gain, people need to be aware of the flag and use it. There is the option to pass `max` to `node-gyp` which will set jobs to the number of cores available. If this was the default for `electron-rebuild` the performance boost would be available out of the box.

@MarshallOfSound, thoughts?

PR #1106, opened 2023-10-01
feat: add support for node-gyp --jobs option to parallelize an indivi…

nkallen:
Hi,

I have a native module with a couple hundred c++ files. Electron-rebuild does have a "parallel" option but this doesn't actually run make in parallel WITHIN one module. This adds support for node-gyp's --jobs option, so that users like me can run electron-rebuild --jobs=max from the command line

Thanks


nkallen:
Hi,

Are you guys not accepting submissions? I'm pretty
Sure this is a useful feature for everyone

BlackHole1 (approved):
LGTM. Although it can be achieved through the [JOBS environment variable](https://github.com/nodejs/node-gyp/blob/b3d41aeb737ddd54cc292f363abc561dcc0a614e/lib/build.js#L27) in node-gyp, I personally feel that supporting the `--jobs` parameter would be better. Thank you for your contribution!

PTAL @dsanders11

erickzhao:
ref https://github.com/electron/rebuild/issues/303

dsanders11:
References #303.

README.md
```diff
@@ -72,6 +72,7 @@ Options:
                                Default is "prod,optional"
   -p, --parallel               Rebuild in parallel, this is enabled by default
                                on macOS and Linux
+  -j, --jobs                   Run make in parallel. The value max will use all available CPU cores
```
erickzhao, line 75:
```suggestion
  -j, --jobs                   Run make in parallel. The value max will use all available CPU cores. Defaults to 1.
```

src/cli.ts
```diff
@@ -26,6 +26,7 @@ const argv = yargs(process.argv.slice(2)).version(false).options({
   types: { alias: 't', type: 'string', description: 'The types of dependencies to rebuild.  Comma separated list of "prod", "dev" and "optional".  Default is "prod,optional"' },
   parallel: { alias: 'p', type: 'boolean', description: 'Rebuild in parallel, this is enabled by default on macOS and Linux' },
   sequential: { alias: 's', type: 'boolean', description: 'Rebuild modules sequentially, this is enabled by default on Windows' },
+  jobs: {alias: 'j', type: 'string', description: 'Run make in parallel. The value max will use all available CPU cores'},
```
erickzhao, line 29:
```suggestion
  jobs: {alias: 'j', type: 'string', description: 'Run make in parallel. The value max will use all available CPU cores. Defaults to 1.'},
```

erickzhao (approved):
Small docs change, but LGTM

dsanders11 (changes requested):
I think this may be currently possible by setting the `npm_config_jobs` environment variable, [see the `node-gyp` documentation](https://github.com/nodejs/node-gyp#environment-variables).

@nkallen, could you test using the environment variable to change the number of jobs and see if that works for your use case? If it does, I'd rather document in this repo that you can pass additional options to `node-gyp` in this manner, rather than plumbing in through. Someone may come along with the need for another `node-gyp` option, like `--python`, and plumbing them through on a case-by-case basis doesn't seem ideal if they're easy to set via environment variables.

dsanders11:
Closing as this was superseded by #1290.

PR #1114, opened 2023-11-18
ci: bump continuousauth/npm orb to 2.1.0

dsanders11:
As in title.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  cfa: continuousauth/npm@1.0.2
+  cfa: continuousauth/npm@2.1.0
   node: electronjs/node@1.2.0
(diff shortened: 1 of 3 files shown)


VerteDinde (approved):

dsanders11:
This turned out to be more annoying than intended. Removing `@continuous-auth/semantic-release-npm` unexpectedly caused CI failures on Node.js v12 and v14. Turns out those were working because the transitive dependencies for that package installed `npm@^8.3.0` which ended up being used in the tests. That silently worked around an issue where the latest Python 3 versions don't work with older npm versions.

Fixed by updating npm on the Node.js v12 and v14 jobs, but it's a headache on Windows so I punted on that and simply excluded those from the matrix, not worth the headache.

erikian (approved):

PR #1115, opened 2023-11-18
fix: use 'target_arch' for node-pre-gyp

samuelmaddock:
The arguments provided to node-pre-gyp are incomplete, resulting in prebuilt binaries not downloading the correct architecture. [From the readme](https://github.com/mapbox/node-pre-gyp#options), `--target_arch` and `--target_platform` are both used for downloading the prebuilt binaries. The `--arch` option seems to only be used when falling back to the `node-gyp` local build.

Also, if the binary already exists, `--update-binary` is required to redownload in the case of changing the targeted architecture. 

Follow up to https://github.com/electron/rebuild/pull/1095

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "node-api-version": "^0.1.4",
     "node-gyp": "^9.0.0",
     "ora": "^5.1.0",
-    "read-binary-file-arch": "^1.0.5",
+    "read-binary-file-arch": "^1.0.6",
     "semver": "^7.3.5",
     "tar": "^6.0.5",
     "yargs": "^17.0.1"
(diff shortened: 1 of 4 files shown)


MarshallOfSound (approved):

samuelmaddock:
[aea4654](https://github.com/electron/rebuild/pull/1115/commits/aea4654d3f1bcfb75b50a46509a5b954dde533cc) will unfortunately cause the binaries to be redownloaded each time a rebuild occurs. node-pre-gyp doesn't update the binary without the `--update-binary` option even if the architecture changes.

MarshallOfSound:
@samuelmaddock is there a way that we can be smarter here for them? E.g. detect a .node file with the wrong arch and nuke it

Not super fussed but would be nice if it's not too hard

MarshallOfSound (approved):

Issue #1116, opened 2023-11-26
need to upgrade node-gyp


soroushm:
[distutils has been removed in Python v3.12](https://docs.python.org/3/whatsnew/3.12.html#summary-release-highlights). we need to upgrade version of node-gyp (to v10+)
```
› Traceback (most recent call last):
  File "xxxxx\node_modules\@electron\rebuild\node_modules\node-gyp\gyp\gyp_main.py", line 42, in <module>
  import gyp  # noqa: E402
  ^^^^^^^^^^
  File "xxxxxx\node_modules\@electron\rebuild\node_modules\node-gyp\gyp\pylib\gyp\__init__.py", line 9, in <mod…
  import gyp.input
  File "xxxxx\node_modules\@electron\rebuild\node_modules\node-gyp\gyp\pylib\gyp\input.py", line 19, in <modul…
  from distutils.version import StrictVersion
  ModuleNotFoundError: No module named 'distutils'
  Error: `gyp` failed with exit code: 1
      at ChildProcess.onCpExit (xxxx\node_modules\@electron\rebuild\n…

An unhandled rejection has occurred inside Forge:
Error: node-gyp failed to rebuild 'xxxxx2\node_modules\robotjs'
at ChildProcess.<anonymous> (xxxxx\node_modules\@electron\rebuild\lib\module-type\node-gyp\node-gyp.js:118:24)
    at ChildProcess.emit (node:events:537:28)
    at ChildProcess._handle.onexit (node:internal/child_process:291:12)
```

si458:
can confirm same issue, a temp fix is to include the latest `node-gpy` as a devDependency in YOUR package.json
`npm install node-gyp --save-dev` should install 10.0.1

aquacode:
Hi,
I'm struggling to get electron-rebuild to work due to node-gyp version and Python version (as others reported above).

I've tried the so-called temp fix by adding the devDependency in my package.json to the latest node-gyp but it's not getting used by electron-rebuild.

Is there any other known workaround for this issue?

tarekis:
As references in https://github.com/nodejs/node-gyp/issues/2869#issuecomment-1966234393, node-gyp v10+ solves this error; distutils was removed from python 3.12 as mentioned [here](https://stackoverflow.com/questions/69919970/no-module-named-distutils-but-distutils-installed/76691103#76691103), which causes this error.

[...]

PR #1128, opened 2024-03-02
fix!: update node-gyp to ^10.0.0 and fix subsequent breaking changes

tarekis:
I took a shot at upgrading `node-gyp` to `^10.0.0`, because electron-rebuild does not work with python `3.12.x` as mentioned in https://github.com/electron/rebuild/issues/1116.
This change is non-breaking and backwards-compatible with python `3.11.x` since `node-gyp@10.0.1` also is.

The obvious change here is updaring `node-gyp` to  `^10.0.0`, but there are some errors that arise from doing so.

So this PR also fixes errors introduced by this upgrade:

- Removes promisifying of node-gyp commands called in the worker, and instead calls the node-gyp APIs directly as they were promisified internally as notend in the `v10.0.0` [breaking changes](https://github.com/nodejs/node-gyp/releases/tag/v10.0.0).

- _dev only:_ `@types/minipass` are set to `^3.3.5` because tar has a dependency version of * but the set version is required, as recommended [here](https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/60901#discussioncomment-3217607).

- _dev only:_ Trough peer dependencies `v10.0.0` introduces dependencies to `string-width` @ `^5.0.1` and `^5.1.2`, which are ESM modules, and since `@electron/rebuild` targets CommonJS this fails with [ESM Errors](https://github.com/TypeStrong/ts-node#err_require_esm). Since a yarn lockfile is used, I used yarn-specific [overrides](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides) (speak: [resolutions](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/)) to force the usage of `4.2.3`, which does not target ESM, and thus works with a CommonJS target.
<sub>This probably could be done more gracefully, but this works for now, and only affects dev, so if you'd like to have this done differently, please manually seek some versions that work and update the lockfile.</sub>


tarekis:
I suppose it is breaking for node v12/14.
What is the policy here, considering these versions are officially out of maintenance?

Also, this doesn't seem to work on Windows. Well, that's a bummer. 

This fails:
```
C:\Users\circleci\AppData\Local\Temp\electron-rebuild-test\node_modules\ffi-napi\build\deps\libffi\ffi.targets(34,5): error MSB3721: The command "call "call" "../../../deps/libffi/preprocess_asm.cmd" "../../../deps/libffi/include" "../../../deps/libffi/config/win/x64" "..\..\..\deps\libffi\src\x86\win64_intel.preasm" "Release\obj\ffi\win64_intel.asm"" exited with code 1. [C:\Users\circleci\AppData\Local\Temp\electron-rebuild-test\node_modules\ffi-napi\build\deps\libffi\ffi.vcxproj]
Done Building Project "C:\Users\circleci\AppData\Local\Temp\electron-rebuild-test\node_modules\ffi-napi\build\deps\libffi\ffi.vcxproj" (default targets) -- FAILED.
Done Building Project "C:\Users\circleci\AppData\Local\Temp\electron-rebuild-test\node_modules\ffi-napi\build\binding.sln" (default targets) -- FAILED.
```

I'm not experience enough to comment on why specifically windows won't build here as of yet. Maybe someone else has a clue.

Latest [cricleci before this change](https://app.circleci.com/pipelines/github/electron/rebuild/1738/workflows/21db4118-633f-4852-bdc9-3568f5507351/jobs/10383?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary) suceeds for Windows + Node 20, [circli for this change](https://app.circleci.com/pipelines/github/electron/rebuild/1743/workflows/d2773c06-4060-4fea-bfb2-af506575001d/jobs/10447?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary) does not.

MarshallOfSound (commented):
`resolutions` won't propagate to folks who install this module, this would be a breaking change for them.  This is on my list of things to look at

clemenshimmer:
I see, the `resolutions` introduced by me [here](https://github.com/electron/rebuild/pull/1128/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R84) do not propagate when one installs?
Was I mistaken assuming this only affects developing `@electron/rebuild`, and also affects the actual usage of `electron-rebuild` when installed?

Totally broken on Windows still tho, no idea what fails on the Win build system tho. IMHO node v12/14 could realistically be wontfixes, but Windows has to work first anyway.

erickzhao:
Hey folks, I just merged `main` back onto this branch to fix merge conflicts. I thought Windows CI being broken would be fixed by pulling in https://github.com/electron/rebuild/pull/1145 but it looks like that isn't the case.

> I suppose it is breaking for node v12/14.
What is the policy here, considering these versions are officially out of maintenance?

We just defined a new policy for `engines` requirements for our various `@electron/` packages: https://github.com/electron/governance/tree/main/wg-ecosystem#nodejs-upgrade-policy

We're looking to make a major version bump in lockstep with all our packages sometime in the future for Node.js 20 so that we can avoid ecosystem fragmentation.

clemenshimmer:
Thanks for the update @erickzhao !

I haven't gotten around to look further into why Windows doesn't build, and I honestly don't get along well with C-anything building, especially on Windows. Really hoped node-gyp would just cover that. Bummer.

Good on the upgrade policy! Now Windows should just work...

<sub>PS: Tarekis and me are the same person, just a different user, sorry if that's confusing.</sub>

PR #1141, opened 2024-06-11
docs: update option documentation

erickzhao:
Adds inline documentation for options and updates the CLI options according to the latest `--help` flag output.

Part of the push to add TypeDoc everywhere but stops just short since there are Node.js compatibility issues with the latest versions of TypeDoc.


README.md
```diff
@@ -54,33 +54,44 @@ installation/runtime requirements.
...
+                                 Default is "prod,optional"             [string]
+  -p, --parallel                 Rebuild in parallel, this is enabled by default
+                                 on macOS and Linux                    [boolean]
+  -s, --sequential               Rebuild modules sequentially, this is enabled
+                                 by default on Windows                 [boolean]
+  -b, --debug                    Build debug version of modules        [boolean]
+      --prebuild-tag-prefix      GitHub tag prefix passed to prebuild-install.
+                                 Default is "v"                         [string]
+      --force-build-from-source  Skip prebuild download and rebuild module from
+                                 source. Default is false              [boolean]
```
erikian, line 86: issue: I think we should match the argument name that we have in `RebuildOptions`, the CLI expects an argument called `buildFromSource` so passing `--force-build-from-source=true` has no effect right now: https://github.com/electron/rebuild/blob/a93581c57d7bf9f1408693fc40614ccc10f51488/src/cli.ts#L131

erikian (changes requested):

README.md
erickzhao: 🤔 Just regenerated those based on the output from `node lib/cli.js --help`

dsanders11 (approved):

README.md
erikian: ah `buildFromSource` is not listed as an option here, so it doesn't show up in the help output:

https://github.com/electron/rebuild/blob/b8ac6dbb3eed2a133b41700ab36940c28e9264e3/src/cli.ts#L14-L34

Since we do list `buildFromSource` in `RebuildOptions`, I think we just forgot to add it to the CLI options when it was first implemented?

erickzhao: Good point: [ae9de35](https://github.com/electron/rebuild/pull/1141/commits/ae9de351f9f413ec1b60ea3ef82b852ead63dddb)

erikian (approved):

PR #1148, opened 2024-08-12
chore(deps): replace ora and chalk with leaner alternatives

Namchee:
## Overview

This PR replaces `ora` and `chalk` in favor of `nanospinner` and `picocolors` without changing the current behavior of the CLI while reducing its size.

[Current dependency graph](https://npmgraph.js.org/?q=@electron/rebuild)

### Package Report

Screenshot 2024-08-12 213710
Screenshot 2024-08-12 213823


erickzhao (changes requested):
Hey @Namchee, thank you for taking the time to raise this PR! After some discussion with other maintainers, we've decided to decline this PR.

Our rationale is that bundle size alone is not a strong enough reason for replacing well-known packages that have a history of stability and consistent maintenance.

cc @electron/wg-ecosystem

PR #1153, opened 2024-09-24
fix: cross-platform prebuild downloads

mmaietta:
Allow prebuilt binaries to be downloaded for target platform instead of hardcoded `process.platform`. Throws an error when attempting to cross-compile (`buildFromSource` || no prebuilds found)
Example: https://github.com/electron/rebuild/blob/cb372cdf09d9ca3ac8f1231608e4ebedddabe51c/src/module-type/node-pre-gyp.ts#L31

Origin of issue: https://github.com/electron/rebuild/blob/cb372cdf09d9ca3ac8f1231608e4ebedddabe51c/src/rebuild.ts#L135

Updated interface with new `platform` option and added docs

### DIFF
--- a/src/module-rebuilder.ts
+++ b/src/module-rebuilder.ts
@@ -151,10 +151,6 @@ export class ModuleRebuilder {
       return true;
     }
 
-    if (this.rebuilder.platform !== process.platform) {
-      throw new Error("OS Limitation - It is not possible to cross-compile native modules from source")
-    }
-
     return await this.rebuildNodeGypModule(cacheKey);
   }
 }
--- a/src/module-rebuilder.ts
+++ b/src/module-rebuilder.ts
@@ -151,6 +151,10 @@ export class ModuleRebuilder {
       return true;
     }
 
+    if (this.rebuilder.platform !== process.platform) {
+      throw new Error("OS Limitation - It is not possible to cross-compile native modules from source")
+    }
+
     return await this.rebuildNodeGypModule(cacheKey);
   }
 }
--- a/src/rebuild.ts
+++ b/src/rebuild.ts
@@ -20,6 +20,13 @@ export interface RebuildOptions {
    */
   electronVersion: string;
   /**
+   * Override the target platform to something other than the host system platform.
+   * Note: This only applies to downloading prebuilt binaries. **It is not possible to cross-compile native modules**
+   * 
+   * @defaultValue The system {@link https://nodejs.org/api/process.html#processplatform | `process.arch`} value
+   */
+  platform?: string;
+  /**
    * Override the target rebuild architecture to something other than the host system architecture.
    * 
    * @defaultValue The system {@link https://nodejs.org/api/process.html#processarch | `process.arch`} value
@@ -132,7 +139,7 @@ export class Rebuilder implements IRebuilder {
   public lifecycle: EventEmitter;
   public buildPath: string;
   public electronVersion: string;
-  public platform: string = process.platform;
+  public platform: string
   public arch: string;
   public force: boolean;
   public headerURL: string;
@@ -151,6 +158,7 @@ export class Rebuilder implements IRebuilder {
     this.lifecycle = options.lifecycle;
     this.buildPath = options.buildPath;
     this.electronVersion = options.electronVersion;
+    this.platform = options.platform || process.platform
     this.arch = options.arch || process.arch;
     this.force = options.force || false;
     this.headerURL = options.headerURL || 'https://www.electronjs.org/headers';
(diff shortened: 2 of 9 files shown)


src/module-rebuilder.ts
```diff
@@ -151,6 +151,10 @@ export class ModuleRebuilder {
       return true;
     }
 
+    if (this.rebuilder.platform !== process.platform) {
+      throw new Error("OS Limitation - It is not possible to cross-compile native modules from source")
+    }
```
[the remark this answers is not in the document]
mmaietta: I took a look at that approach, but the reason I didn't was because then it seemed like electron-rebuild would try to rebuild every native module and throw an individual error each time
https://github.com/electron/rebuild/blob/cb372cdf09d9ca3ac8f1231608e4ebedddabe51c/src/rebuild.ts#L237-L241

Instead, I felt that putting the check at the beginning would be an early exit from the flow

src/rebuild.ts
```diff
@@ -19,6 +19,13 @@ export interface RebuildOptions {
    * The version of Electron to build against.
    */
   electronVersion: string;
+  /**
+   * Override the target platform to something other than the host system platform.
+   * Note: This only applies to downloading prebuilt binaries. **It is not possible to cross-compile native modules**
```
erickzhao, line 24:
```suggestion
   * Note: This only applies to downloading prebuilt binaries. **It is not possible to cross-compile native modules.**
```

Tiny punctuation nit but otherwise LGTM

erickzhao (approved):

mmaietta:
Writing a few tests for the code coverage requirement. Stay tuned.

mmaietta:
Good thing I added tests. I had missed adding `platform` to the cache key.

**I could use a tad bit of help though.** (I'll continue investigating as well) I'm running into this test failure for the "early exit" when attempting cross-compilation via node-gyp.
```
  1 failing

  1) node-gyp
       buildArgs
         cross-compilation
           throws error early:
     AssertionError: expected [Function executor] to throw an error
```
Oddly enough, if I `console.log` right before the `throw new Error`, it logs the state correctly. So for some reason, `expect(executor).throws()` is not receiving the `Error` being thrown. Any idea why?
https://github.com/electron/rebuild/blob/fff22551593b4f41a51d8420164a4233d0b47c44/src/module-type/node-gyp/node-gyp.ts#L86-L89
https://github.com/electron/rebuild/blob/fff22551593b4f41a51d8420164a4233d0b47c44/test/module-type-node-gyp.ts#L69-L74

Also, quick note, I changed `platform: string` to `platform: NodeJS.Platform` which shouldn't change much for the end user since `platform` was never configurable beforehand, so it just helps intellisense out for end users now.

src/module-rebuilder.ts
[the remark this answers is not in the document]
mmaietta: Following up here, I ended up moving the logic into `rebuildModule` as you suggested to simplify/isolate behaviors for unit testing specific rebuild/prebuild files as opposed to pulling the entire `ModuleRebuilder` into `module-type-node-gyp.ts` test case

mmaietta:
Investigation update. It's throwing the error as written, but the event isn't propagating to mocha specifically. It properly logs the error being thrown if I add a `.catch(console.error)` to `nodeGyp.rebuildModule()`. I've taken a different approach in using an error message equality (to make sure we're testing this specific error being thrown) and then `expect` on a boolean flag

All tests are passing 🙃 

https://github.com/electron/rebuild/blob/0fbaba08cc7055f394a3ad231ee630d2cbb6b663/test/module-type-node-gyp.ts#L75-L82

MarshallOfSound:
@mmaietta `throws` is used for synchronous checks, because the function you're calling is async you need to use the `eventually` assertion variants.

E.g. `.to.eventually.be.rejectedWith` (there are examples of this in other tests)

mmaietta:
Thanks @MarshallOfSound! I had overlooked the `chaiAsPromised`. Updated the test accordingly

src/rebuild.ts
```diff
@@ -151,6 +158,7 @@ export class Rebuilder implements IRebuilder {
     this.lifecycle = options.lifecycle;
     this.buildPath = options.buildPath;
     this.electronVersion = options.electronVersion;
+    this.platform = options.platform || process.platform
```
MarshallOfSound, line 161: Missing a semicolon, surprised lint didn't complain

mmaietta: 👍 
Double-checked all new lines for semicolons and added where I had missed them. PR updated.

mmaietta: @MarshallOfSound I've opened a separate PR for fixing the semicolon lint issue 🙂 
https://github.com/electron/rebuild/pull/1154

samuelmaddock (approved):
thanks for the added tests!

mmaietta:
😞 Looks like this needed to be rebased off latest `main` since the eslint PR went in (or vice versa). My apologies

PR #1154, opened 2024-09-25
chore: enforce semicolons via eslint

mmaietta:
Adding eslint rule to enforce semicolons and fixing related lines/errors

Created this PR due to this comment: https://github.com/electron/rebuild/pull/1153#discussion_r1775646121

Note: `"semi": "off"` is needed as `"@typescript-eslint/semi": "error",` supersedes `semi` and having both as `error` results in duplicated errors for each lint line. Ref: https://stackoverflow.com/a/72341604/3498974

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,10 @@
           "ignoreRestSiblings": true,
           "vars": "all"
         }
-      ]
+      ],
+      "semi": "off",
+      "@typescript-eslint/semi": "error",
+      "no-unexpected-multiline": "error"
     }
   },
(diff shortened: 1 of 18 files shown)


malept:
Seems like a chore and not a fix (doesn't require a new release)

erickzhao (approved):
Interesting that this repo doesn't have our standard Prettier + ESLint setup but I think this is a good change without having to add another dependency.

samuelmaddock (approved):

dsanders11 (approved):

PR #1155, opened 2024-10-01
fix: allow using forks for prebuilt modules

samuelmaddock:
It may sometimes be necessary to fork a prebuilt module package to fix an issue. In such cases, the rebuilder can miss prebuilt modules, forcing them to be built from source.

To avoid this, the module rebuilders will now also search for possible forked packages.
- `@mapbox/node-pre-gyp` may instead be `node-pre-gyp` or `@namespace/node-pre-gyp`
- `prebuild-install` may instead be `@namespace/prebuild-install`
- `prebuilds` may instead be `@namespace/prebuilds`

### DIFF
--- a/src/module-type/index.ts
+++ b/src/module-type/index.ts
@@ -58,16 +58,18 @@ export class NativeModule {
    * Search dependencies for package using either `packageName` or
    * `@namespace/packageName` in the case of forks.
    */
-  async findPackageInDependencies(packageName: string): Promise<string | null> {
-    const dependencies = await this.packageJSONFieldWithDefault('dependencies', {});
+  async findPackageInDependencies(packageName: string, packageProperty = 'dependencies'): Promise<string | null> {
+    const dependencies = await this.packageJSONFieldWithDefault(packageProperty, {});
     if (typeof dependencies !== 'object') return null;
-    for (const dependency of Object.keys(dependencies)) {
-      // Directly using package
-      if (dependency === packageName) return dependency;
-      // Using possible forked package
-      if (dependency.endsWith(`/${packageName}`)) return dependency;
-    }
-    return null;
+
+    // Look for direct dependency match
+    // eslint-disable-next-line no-prototype-builtins
+    if (dependencies.hasOwnProperty(packageName)) return packageName;
+
+    const forkedPackage = Object.keys(dependencies).find(dependency =>
+      dependency.startsWith('@') && dependency.endsWith(`/${packageName}`));
+
+    return forkedPackage || null;
   }
 }
 
--- a/src/module-type/index.ts
+++ b/src/module-type/index.ts
@@ -53,6 +53,22 @@ export class NativeModule {
 
     return binary?.napi_versions;
   }
+
+  /**
+   * Search dependencies for package using either `packageName` or
+   * `@namespace/packageName` in the case of forks.
+   */
+  async findPackageInDependencies(packageName: string): Promise<string | null> {
+    const dependencies = await this.packageJSONFieldWithDefault('dependencies', {});
+    if (typeof dependencies !== 'object') return null;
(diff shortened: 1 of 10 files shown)


VerteDinde (approved):

MarshallOfSound (changes requested):
I feel like this is gonna trip false positives, especially on `@namespace/prebuilds`. This is a massive edge case I'd rather cause slightly more work for folks doing this than Just Guess And Hope.

If we can come up with a good heuristic for identifying forks beyond just "is the suffix still the same" I'd be down.

E.g. Maybe we can check other fields in the suspect `package.json` such as `repo` or something and see if it matches the upstream repo url.

Either that or letting modules provide a field in the package.json that we specifically read and understand.  E.g. `packageJSON.electronRebuildConfig = { moduleType: 'prebuilds', builderModuleName: '@foo/prebuilds' }`

samuelmaddock:
> I feel like this is gonna trip false positives, especially on @namespace/prebuilds

The nice thing about our current architecture is that even making the wrong assumption will fallback gracefully.
https://github.com/electron/rebuild/blob/9ebf9bd51c8e602f19fbd39ed01b21e9e60096b5/src/module-rebuilder.ts#L144-L155

In each case, even if `@namespace/prebuildify` is a false positive, `findPrebuiltModule()` should fail to find the module unless the APIs match exactly. Once it fails, it can fallthrough to building from source.
https://github.com/electron/rebuild/blob/9ebf9bd51c8e602f19fbd39ed01b21e9e60096b5/src/module-rebuilder.ts#L66-L78

Looking through the public packages for [node-pre-gyp](https://www.npmjs.com/search?q=node-pre-gyp&ranking=popularity), [prebuildify](https://www.npmjs.com/search?ranking=popularity&q=prebuildify), and [prebuild-install](https://www.npmjs.com/search?ranking=popularity&q=prebuild-install) I see no false positives that would cause any fatal error.

> If we can come up with a good heuristic for identifying forks beyond just "is the suffix still the same" I'd be down.
> 
> E.g. Maybe we can check other fields in the suspect package.json such as repo or something and see if it matches the upstream repo url.

I was considering this when looking through [node-pre-gyp forks sorted by popularity on npmjs.](https://www.npmjs.com/search?q=node-pre-gyp&ranking=popularity) Enough forks change the repo URL to point to their own fork where this probably wouldn't be reliable enough.

Using the `bin` seems to be the most reliable indicator. This is already covered as we search for the binary and skip if it's not found (implemented similarly in each prebuilt module type).
https://github.com/electron/rebuild/blob/f182a4f942f783a972998af10674c9d37da28aa8/src/module-type/node-pre-gyp.ts#L14-L21
https://github.com/electron/rebuild/blob/f182a4f942f783a972998af10674c9d37da28aa8/src/module-type/node-pre-gyp.ts#L44-L46

> Either that or letting modules provide a field in the package.json that we specifically read and understand. E.g. `packageJSON.electronRebuildConfig = { moduleType: 'prebuilds', builderModuleName: '@foo/prebuilds' }`

This is a good way to eliminate the possibility of false positives, but requires more work for folks which I'd hopefully like to avoid.

-----

In summary, I'm less concerned by false positives since
- rebuild gracefully fallsback to building from source if each of the prebuilt modules fail
- rebuild already searches for the exact binary required by the prebuilt modules (e.g. `node_modules/@namespace/node-pre-gyp/bin/node-pre-gyp`)
- the names for prebuilt modules (prebuildify, prebuild-install, node-pre-gyp) are fairly unique
- none of the public `@namespace/{prebuildify|prebuild-install|node-pre-gyp}` modules on npm indicate that they would trip up our rebuild logic

MarshallOfSound (approved):
Agreed that given other code stars would have to align to false-positive. CI is still broken, but approving the concept

PR #1157, opened 2024-10-07
feat: Change node-gyp to @electron/node-gyp

### LINKED ISSUE #1116: need to upgrade node-gyp

felixrieseberg:
This PR switches `@electron/rebuild` to a fork of the original and excellent `node-gyp` with only one feature added: Support for Node v12. We're doing so to fix `ModuleNotFoundError: No module named 'distutils'` errors on macOS Sequoia, which ships with Python 3.12 by default. Here is what happened:

- Python 3.12 has removed `distutils`, which `node-gyp` below v10 depended on.
- macOS Sequoia updated their default version of Python to 3.12.
- When installing a native Node addon, a lot of code bases suddenly showed `ModuleNotFoundError: No module named 'distutils'` error messages that are cryptic for anyone who doesn't have a full understanding of the dependency chain involved

This left us with the following choices:

1. **Tell the community to manually install either Python 3.11 or `setuptools`**. While this fixes the issue, it requires manual user intervention and requires our users to first search for the error message.
2. **Upgrade to `node-gyp` v10**. That would have required a bump in minimum Node.js version for `@electron/rebuild` and in turn all other packages that depend on it, resulting in major version bumps across the entire ecosystem. This too doesn't fix the issue for our users without them performing manual major version upgrades.
3. **Fork node-gyp v10, make it compatible with Node.js v12, and only patch our tools**: We chose this option to get a fix out to our users as quickly and painless as possible.

Electron has a major version bump in required Node.js version on its roadmap. Once that is the case, we will sunset this fork and go back to the original `node-gyp`.

closes #1116

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "node": ">=12.13.0"
   },
   "dependencies": {
-    "@electron/node-gyp": "^10.2.0-electron.1",
+    "@electron/node-gyp": "https://github.com/electron/node-gyp#fd1e8912a0a0b1d839bba9046fbeb60fa2b0d76b",
     "@malept/cross-spawn-promise": "^2.0.0",
     "chalk": "^4.0.0",
     "debug": "^4.1.1",
--- a/package.json
+++ b/package.json
@@ -29,7 +29,10 @@
   "keywords": [
     "electron"
   ],
-  "author": "Paul Betts <paul@paulbetts.org>",
+  "authors": [
+    "Ani Betts <anais@anaisbetts.org>",
+    "Electron Community"
+  ],
   "license": "MIT",
   "bugs": {
     "url": "https://github.com/electron/rebuild/issues"
@@ -39,6 +42,7 @@
     "node": ">=12.13.0"
   },
   "dependencies": {
+    "@electron/node-gyp": "^10.2.0-electron.1",
     "@malept/cross-spawn-promise": "^2.0.0",
     "chalk": "^4.0.0",
     "debug": "^4.1.1",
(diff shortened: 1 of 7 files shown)


samuelmaddock (approved):
Thanks for taking this on @felixrieseberg! I just ran into this last week 🙇

Toinane (approved):

PR #1166, opened 2025-01-16
build(dev-deps): bump mocha to clear the audit

dsanders11:
This clears the audit. The new version of `mocha` requires Node.js >= 14 so this also drops Node.js 12 from our CI matrix - we'll be doing the Node.js minimum bump soon enough.

### DIFF
--- a/test/helpers/module-setup.ts
+++ b/test/helpers/module-setup.ts
@@ -39,7 +39,7 @@ export async function resetTestModule(testModulePath: string, installModules = t
 }
 
 export async function cleanupTestModule(testModulePath: string): Promise<void> {
-  await fs.rmdir(testModulePath, { recursive: true, maxRetries: 10 });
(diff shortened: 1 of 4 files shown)


BlackHole1 (approved):

test/helpers/module-setup.ts
```diff
@@ -39,7 +39,7 @@ export async function resetTestModule(testModulePath: string, installModules = t
 }
 
 export async function cleanupTestModule(testModulePath: string): Promise<void> {
-  await fs.rmdir(testModulePath, { recursive: true, maxRetries: 10 });
+  await fs.rm(testModulePath, { recursive: true, maxRetries: 10 });
```
dsanders11, line 42: Clears a warning that was showing up during test runs about `fs.rmdir({ recursive: true })` being deprecated.

PR #1167, opened 2025-01-16
feat!: bump `engines` requirement to Node 22

erikian:
BREAKING CHANGE: bumps required Node.js version to `>=22.12.0`. ESM-only.

I also took the opportunity to update `farmhash` to `3.3.1` in one of the test fixtures so we can run a couple of test cases we were skipping on `arm64` macOS. I tried switching to `4.0.1` but it would require some deeper changes in the tests, so I figured it would be a good idea to address that in a follow-up PR as `3.3.1` is enough to get those tests running for the Node upgrade.


felixrieseberg (approved):
Looks clean and good!

.eslintignore
```diff
@@ -0,0 +1,2 @@
+*.d.ts
```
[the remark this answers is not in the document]
erikian: I think ESLint tries to lint `.d.ts` files by default because some projects have them in version control to handle global types, but in our case it makes sense to skip all the generated declaration files that live in the `lib/` folder.

README.md
```diff
@@ -155,7 +155,7 @@ const childProcess = require('child_process');
 const pathToElectron = require('electron');
 
   rebuild({
-    buildPath: __dirname,
+    buildPath: import.meta.dirname,
```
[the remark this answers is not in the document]
erikian: 06b36a23a7a71588b1e0206b3c7c1d6ff8066f82

README.md
```diff
@@ -163,4 +164,18 @@ const pathToElectron = require('electron');
       console.error("Building modules didn't work!");
       console.error(e);
     });
+  
+// ESM
+import childProcess from 'child_process';
+import pathToElectron from 'electron';
```
erickzhao, line 170: What did this `pathToElectron` import do in this example? Weird variable name for the import.

package.json
```diff
@@ -2,17 +2,17 @@
...
+  "exports": "./lib/main.js",
   "typings": "lib/main.d.ts",
   "scripts": {
-    "compile": "tsc",
+    "build": "tsc",
     "coverage": "npm run prewarm-headers && nyc npm run spec",
     "coverage:report": "nyc report --reporter=text-lcov > coverage.lcov",
     "watch": "tsc -w",
-    "prepare": "npm run compile",
-    "mocha": "cross-env TS_NODE_FILES=true mocha",
```
erickzhao, line 13: Was `TS_NODE_FILES` some leftover legacy env var that we didn't need anymore?

src/clang-fetcher.ts
```diff
@@ -47,7 +47,7 @@ export async function getClangEnvironmentVars(electronVersion: string, targetArc
 
   const gypArgs = [];
   if (process.platform === 'win32') {
-    console.log(fs.readdirSync(clangDir));
+    console.log(await fs.promises.readdir(clangDir));
```
erickzhao, line 50: Since this is a `graceful-fs` call, it'll need to be passed through `util.promisify`

src/clang-fetcher.ts
```diff
@@ -102,20 +102,20 @@ async function downloadClangVersion(electronVersion: string) {
...
+
   const contents = await fetch(clangDownloadURL, 'buffer');
   d('deflating clang');
   zlib.deflateSync(contents);
   const tarPath = path.resolve(ELECTRON_GYP_DIR, `${electronVersion}-clang.tar`);
-  if (await fs.pathExists(tarPath)) await fs.remove(tarPath);
-  await fs.writeFile(tarPath, Buffer.from(contents));
-  await fs.mkdirp(clangDirPath);
+  if (fs.existsSync(tarPath)) await fs.promises.rm(tarPath, { recursive: true, force: true });
+  await fs.promises.writeFile(tarPath, Buffer.from(contents));
```
erickzhao, line 111: Since this is a graceful-fs call, it'll need to be passed through util.promisify

src/module-rebuilder.ts
```diff
@@ -120,25 +120,26 @@ export class ModuleRebuilder {
     const buildLocation = path.resolve(this.modulePath, 'build', this.rebuilder.buildType);
 
     d('searching for .node file', buildLocation);
-    const buildLocationFiles = await fs.readdir(buildLocation);
+    const buildLocationFiles = await fs.promises.readdir(buildLocation);
```
erickzhao, line 123: Since this is a `graceful-fs` call, it'll need to be passed through `util.promisify`

src/module-rebuilder.ts
```diff
@@ -120,25 +120,26 @@ export class ModuleRebuilder {
...
-    if (nodePath && await fs.pathExists(nodePath)) {
+    if (nodePath && fs.existsSync(nodePath)) {
       d('found .node file', nodePath);
       if (!this.rebuilder.disablePreGypCopy) {
         const abiPath = path.resolve(this.modulePath, `bin/${this.rebuilder.platform}-${this.rebuilder.arch}-${this.rebuilder.ABI}`);
         d('copying to prebuilt place:', abiPath);
-        await fs.mkdir(abiPath, { recursive: true });
-        await fs.copyFile(nodePath, path.join(abiPath, `${this.nodeGyp.moduleName}.node`));
+        await fs.promises.mkdir(abiPath, { recursive: true });
+        await fs.promises.copyFile(nodePath, path.join(abiPath, `${this.nodeGyp.moduleName}.node`));
```
erickzhao, line 135: Since this is a `graceful-fs` call, it'll need to be passed through `util.promisify`

src/module-rebuilder.ts
```diff
@@ -120,25 +120,26 @@ export class ModuleRebuilder {
...
       }
     }
   }
 
   async writeMetadata(): Promise<void> {
-    await fs.outputFile(this.metaPath, this.metaData);
+    await fs.promises.mkdir(path.dirname(this.metaPath), { recursive: true });
+    await fs.promises.writeFile(this.metaPath, this.metaData);
```
erickzhao, line 142: Since this is a `graceful-fs` call, it'll need to be passed through `util.promisify`

src/read-package-json.ts
```diff
@@ -1,10 +1,12 @@
-import * as fs from 'fs-extra';
-import * as path from 'path';
+import fs from 'graceful-fs';
+import path from 'node:path';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export async function readPackageJson(dir: string, safe = false): Promise<any> {
   try {
-    return await fs.readJson(path.resolve(dir, 'package.json'));
+    return JSON.parse(await fs.promises.readFile(path.resolve(dir, 'package.json'), {
```
erickzhao, line 7: Since this is a `graceful-fs` call, it'll need to be passed through `util.promisify`

src/sysroot-fetcher.ts
```diff
@@ -36,8 +36,8 @@ export async function downloadLinuxSysroot(electronVersion: string, targetArch:
 
   d('writing sysroot to disk');
   const tmpTarFile = path.resolve(ELECTRON_GYP_DIR, `${electronVersion}-${fileName}`);
-  if (await fs.pathExists(tmpTarFile)) await fs.remove(tmpTarFile);
-  await fs.writeFile(tmpTarFile, sysrootBuffer);
+  if (fs.existsSync(tmpTarFile)) await fs.promises.rm(tmpTarFile, { recursive: true, force: true });
+  await fs.promises.writeFile(tmpTarFile, sysrootBuffer);
```
erickzhao, line 40: Since this is a `graceful-fs` call, it'll need to be passed through `util.promisify`

test/search-module.ts
```diff
@@ -22,8 +22,14 @@ describe('search-module', () => {
...
             await createTempDir();
-            await fs.copy(path.resolve(__dirname, 'fixture', 'multi-level-workspace'), baseDir);
-            await fs.ensureFile(path.join(baseDir, lockFile));
+            await fs.promises.cp(path.resolve(import.meta.dirname, 'fixture', 'multi-level-workspace'), baseDir, { recursive: true, force: true });
+
+            const lockfilePath = path.join(baseDir, lockFile);
+
+            if(!fs.existsSync(lockfilePath)) {
+              await fs.promises.mkdir(baseDir, { recursive: true });
+              await fs.promises.writeFile(lockfilePath, Buffer.from([]), {});
```
erickzhao, line 31: Since this is a `graceful-fs` call, it'll need to be passed through `util.promisify`

tsconfig.json
```diff
@@ -1,22 +1,18 @@
...
     "noImplicitThis": true,
     "noUnusedParameters": true,
-    "module": "commonjs",
-    "moduleResolution": "node",
-    "esModuleInterop": true,
     "pretty": true,
-    "target": "ES2019",
-    "outDir": "lib"
+    "outDir": "lib",
+    "types": ["node", "mocha"]
```
erickzhao, line 15: Do we need all these extra tsconfig flags still, or are we able to pare down this list a bit more?

README.md
erikian: Wow, that's leftover code from a [really old](https://github.com/electron/rebuild/commit/20c630fe623bff3ce58de08ae00f137273b46cd1) example. When importing / requiring Electron from a regular Node.js process, [you get the path to the Electron executable](https://github.com/electron/electron/blob/62b5b5f99dad509b9e7bea9ada2e5731d8b04371/npm/index.js#L6-L19) as the default export, and that was used to spawn Electron and get its version.

Since the version is now hardcoded in this example and people would probably use `rebuild` via Forge anyway, I think we can just remove it to keep things simple. I've updated the examples to make them copy-paste-friendly and also bumped `node-abi` and `node-api-version` for good measure - 31dcabd

package.json
erikian: Yes, it was used to tell `ts-node` to load the files included in our `tsconfig.json` before running the tests, but now we're using `tsx` and don't need this anymore.

tsconfig.json
erikian: I removed all flags that are already covered by `@tsconfig/node22`, the ones still left are mostly for added strictness which is nice to have IMO.

src/clang-fetcher.ts
erikian: 6961ea8e01dc73c81112d4d44b9f853e199a8e07

VerteDinde (approved):
Thanks for doing this work!

.github/workflows/docs.yml
```diff
@@ -0,0 +1,35 @@
+name: Publish documentation
+
+on:
+  push:
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+*
+
+permissions:
+      id-token: write
+      contents: read
```
dsanders11, line 10:
```suggestion
  id-token: write
  contents: read
```

dsanders11 (approved):
Minor nit regarding indentation in one of the workflows otherwise LGTM.

erickzhao (approved):
LGTM for what's been addressed from my last review.

PR #1178, opened 2025-04-06
fix(search): improve project root path detection logic

beyondkmp:


src/search-module.ts
```diff
@@ -74,13 +74,17 @@ export async function searchForNodeModules(cwd: string, rootPath?: string): Prom
...
  */
 export async function getProjectRootPath(cwd: string): Promise<string> {
+  let projectRootPath = '';
   for (const lockFilename of ['yarn.lock', 'package-lock.json', 'pnpm-lock.yaml']) {
     const pathGenerator: PathGeneratorFunction = (traversedPath) => path.join(traversedPath, lockFilename);
     const lockPaths = await traverseAncestorDirectories(cwd, pathGenerator, undefined, 1);
     if (lockPaths.length > 0) {
-      return path.dirname(lockPaths[0]);
+      let lockDir = path.dirname(lockPaths[0]);
+      if(lockDir.length > projectRootPath.length) {
```
erickzhao, line 83:
```suggestion
      if (lockDir.length > projectRootPath.length) {
```

src/search-module.ts
```diff
@@ -74,13 +74,17 @@ export async function searchForNodeModules(cwd: string, rootPath?: string): Prom
  * @param cwd the initial directory to traverse
  */
 export async function getProjectRootPath(cwd: string): Promise<string> {
+  let projectRootPath = '';
   for (const lockFilename of ['yarn.lock', 'package-lock.json', 'pnpm-lock.yaml']) {
     const pathGenerator: PathGeneratorFunction = (traversedPath) => path.join(traversedPath, lockFilename);
     const lockPaths = await traverseAncestorDirectories(cwd, pathGenerator, undefined, 1);
     if (lockPaths.length > 0) {
-      return path.dirname(lockPaths[0]);
+      let lockDir = path.dirname(lockPaths[0]);
```
erickzhao, line 82:
```suggestion
      const lockDir = path.dirname(lockPaths[0]);
```

erickzhao (commented):
Can you address the lint failures?

erickzhao (commented):
Can we also add additional test coverage for this use case?

src/search-module.ts
beyondkmp: Fixed.

src/search-module.ts
beyondkmp: Fixed.

beyondkmp:
Yes. UT is added.

PR #1181, opened 2025-04-09
fix: runtime error after removing pnpm module

indutny-signal:
When removing a pnpm module, pnpm might leave a dangling symbolic link to the folder that no longer exists. In this case `realpath` call will fail with ENOENT and fail the build that could otherwise continue without issues.


indutny-signal:
cc @dsanders11 ?

dsanders11 (changes requested):
The `yarn.lock` changes need to be reverted - otherwise code LGTM.

indutny-signal:
@dsanders11 sorry about that. Reverted now!

dsanders11 (approved):

indutny-signal:
Thank you!

PR #1182, opened 2025-04-29
build: bump @electron/rebuild to 4.4.0

VerteDinde:
This PR updates rebuild to use node-abi 4.x.x and higher - node-abi was bumped to 4.x.x during our Node 22 upgrade process, and is now needed to pull in newer versions of Electron.


VerteDinde:
Ah yep, we're going to have to update all of Node, standby

PR #1218, opened 2025-12-09
fix: scoped modules work with `onlyModules` / `--only`

clavin:
Fixes a bug where scoped modules fail when using `onlyModules` config / `--only` flag.

Found out after the fact that this was a TODO in the tests 😄


erickzhao (approved):
:shipit:

PR #1243, opened 2026-03-24
refactor: replace got with native fetch API

MarshallOfSound:
## Summary
Replaced the `got` HTTP client library with the native `fetch` API available in Node.js, removing an external dependency while maintaining the same functionality.

## Key Changes
- Removed `got` import and replaced HTTP requests with `globalThis.fetch()`
- Updated response handling to use the Fetch API's response object:
  - Changed `response.statusCode` to `response.status`
  - Changed `response.body` to conditional handling: `response.arrayBuffer()` for buffer type and `response.text()` for text type
- Removed `got` from package.json dependencies

## Implementation Details
- The fetch API is now natively available in Node.js (v18+), eliminating the need for the external `got` library
- Response type handling is now explicit with conditional logic to convert between `arrayBuffer()` and `text()` based on the requested `responseType` parameter
- The retry logic and error handling remain unchanged

https://claude.ai/code/session_01GSjWK4VdKX3bqqvED6HyXJ

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "@malept/cross-spawn-promise": "^2.0.0",
     "debug": "^4.1.1",
     "detect-libc": "^2.0.1",
-    "got": "^11.7.0",
     "graceful-fs": "^4.2.11",
     "node-abi": "^4.2.0",
     "node-api-version": "^0.2.1",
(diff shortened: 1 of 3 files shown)


dsanders11 (approved):
Worth noting this is going to merge conflict with #1244 which has a slightly different version of this same change.

PR #1244, opened 2026-03-24
fix: minimize production dependency tree

MarshallOfSound:
## Summary

Shrinks the dependency tree by replacing several packages with small inline implementations or Node.js builtins:

| Removed | Replacement |
| --- | --- |
| `debug` | `src/debug.ts` — minimal logger gated on `DEBUG` |
| `detect-libc` | `src/detect-libc.ts` — `process.report`-based glibc/musl check |
| `got` | native `fetch` |
| `graceful-fs` | `node:fs` / `node:fs/promises` |
| `ora` | `src/spinner.ts` — tiny TTY spinner via `util.styleText` |
| `semver`, `tar`, `yargs` | node builtins / simpler equivalents |

Also bumps `node-gyp` to `^12.2.0` and drops the now-unused `src/promisifiedGracefulFs.ts` wrapper along with associated `@types/*` devDeps.

## Test plan

- [ ] `yarn test` passes
- [ ] CLI still parses flags correctly and renders spinner/success/fail states
- [ ] Rebuild against a native module on Linux (glibc + musl) to validate libc detection
- [ ] Cache snapshot/apply roundtrip still works with `node:fs`


dsanders11 (changes requested):
If the goal is just to drop "heavy" dependencies as in the title, I don't like several of these changes.

* `debug` is small and only has one transitive dependency, but we're losing the richer functionality and consistency across our other packages that use `debug`
* `detect-libc` I could take or leave, but it's zero dependency
* `ora` is heavy and we should drop it, but there's zero dependency alternatives rather than rolling our own. We have multiple packages that currently use spinners (most use `ora`), we should migrate them all to the same alternative rather than having each package roll its own implementation that will drift. Or even create `@electron/spinner`, if we're insistent on cutting third-party dependencies.
* We specifically kept `graceful-fs` when we did the Node 22 migration, we should not drop it here.

src/module-type/node-gyp/node-gyp.ts
```diff
@@ -40,7 +37,8 @@ export class NodeGyp extends NativeModule {
     // and --force-process-config must be passed to node-gyp >= 8.4.0 to
     // correctly build modules for them.
     // See also https://github.com/nodejs/node-gyp/pull/2497
-    if (!semver.satisfies(this.rebuilder.electronVersion, '^14.2.0 || ^15.3.0') && semver.major(this.rebuilder.electronVersion) < 16) {
+    const [maj, min] = this.rebuilder.electronVersion.split('.').map(Number);
+    if (maj < 14 || (maj === 14 && min < 2) || (maj === 15 && min < 3)) {
```
erickzhao: atp i'm okay removing the `semver` dep here but i can see the possibility of having to add it back if we ever need trickier conditions.

waiting for semver parsing to be exposed to `node:util` one day i hope 😭

src/cli.ts
```diff
@@ -1,43 +1,47 @@
...
+
+const { values: argv } = parseArgs({ options, allowPositionals: true });
+
+if (argv.help) {
+  console.log('Usage: electron-rebuild --version [version] --module-dir [path]\n\nOptions:');
+  for (const [name, opt] of Object.entries(options)) {
+    const short = 'short' in opt ? `-${opt.short}, ` : '    ';
+    console.log(`  ${short}--${name.padEnd(22)} ${opt.description}`);
+  }
+  console.log('\nCopyright 2016-2021');
```
erickzhao: funny, do we print copyright notices anywhere else in our CLIs? i'd remove this if not tbh

src/clang-fetcher.ts
```diff
@@ -104,18 +102,13 @@ async function downloadClangVersion(electronVersion: string) {
...
-  await promisifiedGracefulFs.writeFile(tarPath, Buffer.from(contents));
+  await fs.promises.writeFile(tarPath, contents);
   await fs.promises.mkdir(clangDirPath, { recursive: true });
-  d('tar running on clang');
-  await tar.x({
-    file: tarPath,
-    cwd: clangDirPath,
-  });
+  d('extracting clang');
+  await spawn('tar', ['-xf', tarPath, '-C', clangDirPath], { stdio: 'ignore' });
```
erickzhao, line 111: just noting here that this moves us from `node-tar` to relying on the system `tar` specifically for this one call, which is specifically used for the advanced `useElectronClang` option.

erickzhao (approved):

src/module-type/node-pre-gyp.ts
dsanders11, line 1: Not gonna block on it, but we really shouldn't have these silly changes in these Claude PRs. This is literally just moving an import around to a different spot, one that makes less sense. Hopefully as part of the `oxc.rs` we can add more linting rules that enforce this stuff.

src/rebuild.ts
dsanders11, line 1: Gonna need an explanation for why these changes are in this PR and why there's a new 'modules-found' lifecycle event being emitted.

dsanders11 (changes requested):

src/rebuild.ts
dsanders11: It looks like it's probably related to dropping `ora`, but I don't think we should have this kind of output change in this PR, as it's unrelated to the stated purpose of the PR.

MarshallOfSound: > Gonna need an explanation for why these changes are in this PR and why there's a new 'modules-found' lifecycle event being emitted.

This isn't new, it was dropped ages ago in a bad refactor and this adds it back

dsanders11: > This isn't new, it was dropped ages ago in a bad refactor and this adds it back

I can't find any evidence of that, at least under the name 'modules-found'. Not gonna block on this, but not sure this is true.

dsanders11 (approved):

